Issue #513, opened 2025-11-29
🔴 BUG: Metadata-Backend Write Inconsistency (No Two-Phase Commit)


windoliver:
## Summary
Write operations lack atomic guarantees between metadata database and backend storage, leading to potential inconsistency on crash.

## Location
`src/nexus/core/nexus_fs_core.py:746+`

## Current Write Sequence
1. Backend writes content (earlier in flow)
2. `metadata.put(metadata)` - writes to DB (line 746)
3. Parent tuples created (line 760-768)
4. Audit log written (line 776-789)

### Scenario A: Crash after metadata write
```
1. ✅ Backend write succeeds
2. ✅ Metadata write succeeds  
3. ❌ CRASH before audit commit
Result: Metadata points to content, but no audit trail
```

### Scenario B: Crash before metadata write
```
1. ✅ Backend write succeeds
2. ❌ CRASH before metadata write
Result: Orphaned content in backend (storage leak)
```

### Scenario C: Partial parent tuple creation
```
1. ✅ Backend write succeeds
2. ✅ Metadata write succeeds
3. ⚠️ Only some parent tuples created
Result: Incomplete permission inheritance
```

## Impact
- Orphaned content blobs (storage leak)
- Metadata pointing to non-existent content
- Incomplete audit trail (compliance violation)
- Broken permission inheritance

## Proposed Fix
1. **Two-Phase Commit Pattern**
   - Phase 1: Write to staging area + prepare metadata
   - Phase 2: Commit metadata + move content to final location

2. **Write-Ahead Log (WAL)**
   - Log intent before operation
   - Replay/rollback on startup

3. **Saga Pattern**
   - Compensating transactions for each step
   - Automatic rollback on failure

## Priority
🔴 High - data consistency issue

## Labels
bug, data-integrity, p0

PR #1695, opened 2026-02-16
fix(#513): change 6 hardcoded zone ID fallbacks from "default" to "root" in subscriptions router

elfenlieds7:
## Summary
- Changed 6 `auth_result.get("zone_id") or "default"` fallbacks to `or "root"` in `server/api/v1/routers/subscriptions.py`
- Aligns with federation-memo.md canonical zone ID (`ROOT_ZONE_ID = "root"`)

## Test plan
- [x] All pre-commit hooks pass (ruff, mypy, format)


joezhoujinjing (commented):
Do we need a default value for zone_id? How about make it explicitly required? @claude

PR #2505, opened 2026-02-22
fix: use getattr for delegation enum value extraction

elfenlieds7:
## Summary
- Fix delegation enum comparison that still fails under pytest-xdist module double-loading
- `isinstance(mode, DelegationMode)` ALSO fails when modules are double-loaded (separate class objects), so the previous fix (PR #2502) didn't fully resolve the issue
- Use `getattr(mode, "value", mode)` to always extract the string value regardless of class identity

## Test plan
- [ ] macOS CI delegation tests pass (test_derivation.py, test_hypothesis.py)
- [ ] Ubuntu CI delegation tests pass
- [ ] All other tests unaffected

### DIFF
--- a/src/nexus/bricks/delegation/derivation.py
+++ b/src/nexus/bricks/delegation/derivation.py
@@ -109,9 +109,9 @@ def derive_grants(
         if existing is None or _relation_rank(relation) > _relation_rank(existing):
             parent_map[object_id] = relation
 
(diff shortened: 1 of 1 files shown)


elfenlieds7:
Closing: this fix was already hotfixed directly to develop (commit 9ff551a5e). The delegation getattr fix is now on develop.

PR #2705, opened 2026-02-28
fix: add rmdir and rename short aliases to RPC dispatch

windoliver:
## Summary
- Add `rmdir` and `rename` short aliases to the RPC dispatch table in `dispatch.py`
- Add corresponding `SysRmdirParams` and `SysRenameParams` entries to `OVERRIDE_METHOD_PARAMS` in `_rpc_param_overrides.py`
- Without these, nexus-test clients using the short method names get "Unknown method" errors

## Test plan
- [x] 55/64 event E2E tests pass (0 failures, 9 infra-dependent skips)
- [x] ruff check + format clean
- [x] mypy passes
- [x] Pre-commit hooks pass


windoliver:
Duplicate of #2703 (same branch, correct target is develop)

Issue #1399, opened 2026-02-12
refactor: Extract Auth module (8,813 LOC) into Auth Brick with Protocol interface


windoliver:
## Summary

Extract the authentication/authorization module from `src/nexus/server/auth/` (8,813 LOC, 23 files, 7+ OAuth providers) into a standalone Auth Brick with a Protocol interface.

## Background

From exhaustive feature audit. The auth module is one of the largest modules NOT covered in the LEGO Architecture doc's brick catalog. It contains:
- 23 files across `src/nexus/server/auth/`
- 7+ OAuth providers (Google, GitHub, Apple, Discord, etc.)
- JWT token management
- API key authentication (`nx_live_*` / `nx_test_*`)
- Session management
- ~8,813 total LOC

This module is tightly coupled to FastAPI server but should be a standalone brick.

## Requirements

1. Define `AuthProtocol` (authenticate, authorize, create_session, etc.)
2. Extract auth logic from server/ into `src/nexus/bricks/auth/`
3. Create `OAuthBaseClass` to reduce duplication across 7 providers (see #1294)
4. Maintain backward compatibility with existing FastAPI middleware
5. Tests: at minimum, port existing tests + add Protocol conformance tests

## References

- `src/nexus/server/auth/` (23 files, 8,813 LOC)
- Related: #1294 (OAuth crypto test suite)
- LEGO Architecture doc (this module was missing from brick catalog)

Issue #1376, opened 2026-02-12
feat(templar): Three-tier plugin discovery (workspace > user > bundled)


windoliver:
## Context

TEMPLAR.md v7.0 Section 1.6.3 defines a three-tier plugin discovery mechanism inspired by OpenClaw's override chain. This enables zero-config defaults with full customizability.

**From:** OpenClaw research (Section 12.12.2) — their 3-tier discovery is the primary reason plugins "just work" for 171K+ users.

### Discovery Chain (highest priority wins)

```
1. Workspace (templar.yaml)     → Project-specific overrides
2. User     (~/.templar/plugins/) → User preferences across projects
3. Bundled  (@templar/*)         → Default implementations
```

### Resolution Algorithm

```typescript
function resolvePlugin(type: string, name: string): TemplarPlugin {
  // 1. Check workspace manifest
  const workspace = manifest.plugins?.[type]?.[name];
  if (workspace) return loadPlugin(workspace);

  // 2. Check user directory
  const userPlugin = scanDir(join(homedir(), '.templar', 'plugins', type, name));
  if (userPlugin) return loadPlugin(userPlugin);

  // 3. Fall back to bundled
  const bundled = require(`@templar/${type}-${name}`);
  if (bundled) return loadPlugin(bundled);

  throw new PluginNotFoundError(type, name);
}
```

### Examples

```yaml
# templar.yaml — workspace overrides bundled Telegram channel
channels:
  telegram:
    package: "./my-custom-telegram"  # Local override
  slack: true                         # Use bundled @templar/channel-slack
middleware:
  my-logger:
    package: "@acme/templar-logger"   # Third-party npm package
```

## Acceptance Criteria

- [ ] Plugin discovery resolves workspace > user > bundled
- [ ] `templar.yaml` can override any bundled plugin
- [ ] `~/.templar/plugins/` directory scanned for user plugins
- [ ] Bundled plugins loaded as fallback
- [ ] Clear error messages when plugin not found at any tier
- [ ] Documentation with examples

## References

- TEMPLAR.md v7.0 Section 1.6.3 (Three-Tier Plugin Discovery)
- TEMPLAR.md v7.0 Section 12.12.2 (Extension Mechanisms Worth Stealing)
- OpenClaw plugin architecture

## Priority: P1 — Depends on #1375 (core/engine split)

PR #2740, opened 2026-03-03
feat: Agent Engine — process model, runtime loop, perf optimizations

windoliver:
## Summary

- Agent Engine Phase 1: full agent process lifecycle (spawn/resume/terminate/signal) with UNIX-inspired FSM
- Agent loop: LLM→tool dispatch→repeat with parallel read-only dispatch and O(log n) context trim
- Tool dispatcher: routes 8 tool calls to NexusFS syscalls and sandbox
- Session store: incremental JSONL checkpoint/restore via CAS
- Queue-based streaming: asyncio.Queue for real-time event delivery
- Architecture design doc + README quick-start section

## Test plan

- [x] 4 E2E tests pass with real Claude Haiku
- [x] ruff + mypy clean
- [ ] CI green


elfenlieds7 (commented):
## Kernel Team Review — Architectural Concerns

Hey Oliver, great work on the Agent Engine! The `system_services/agent_runtime/` code is clean and well-layered. A few kernel-related issues we'd like addressed before merge:

---

### 1. 🔴 `FileEvent.revision` re-introduced in `core/file_events.py` (Blocking)

We deliberately removed `FileEvent.revision` from kernel in #1376 and relocated revision tracking to EventBus service layer in #1382. This PR adds it back:

```python
# core/file_events.py
+ revision: int | None = None
```

Plus `subscriptions.py` and `reactive_subscriptions.py` now read `event.revision or 0`, tying service-layer revision logic back to the kernel struct.

**Ask**: Please revert the `FileEvent.revision` addition and the `subscriptions.py`/`reactive_subscriptions.py` changes. If agent engine needs revision info, it should be sourced from the service-layer event enrichment, not from the kernel `FileEvent` struct. The deleted tests (`TestLazyInit`, `test_wait_for_event_respects_since_revision`) should also be restored.

---

### 2. 🟡 `agent_runtime` in `core/config.py::SystemServices` — wrong layer

`core/config.py` is **kernel** config. Agent runtime is a system service, not a kernel component. Adding it here implies kernel awareness of agent runtime, which breaks the layering:

```python
# core/config.py — SystemServices
+ agent_runtime: Any = None
```

**Ask**: Move the `agent_runtime` field out of `core/config.SystemServices`. If agent runtime needs DI wiring, it should go through the factory layer or a service-layer config, not kernel config.

---

### 3. 🟡 `core/device_capabilities.py` — file was moved to `lib/`

We already moved `device_capabilities.py` from `core/` to `lib/` in PR #2323. Your branch is likely based on an older develop. After rebase, this change should target `lib/device_capabilities.py` instead.

---

### 4. 🟡 `contracts/agent_process.py` + `contracts/deployment_profile.py` placement

`contracts/` is for **kernel-level contracts** that services implement. `AgentProcess`, `AgentProcessConfig`, `AgentEvent`, etc. are agent-engine-specific types, not kernel contracts.

**Ask**: Consider moving `agent_process.py` to `system_services/agent_runtime/contracts.py` (or similar) — keeping agent-engine types co-located with the agent engine. The `BRICK_AGENT_RUNTIME` constant in `deployment_profile.py` is fine (same pattern as other bricks).

---

### 5. 🟢 `contracts/protocols/process_manager.py` — future kernel overlap (tracked)

We're tracking this as a follow-up audit (Task #1399) since kernel will eventually need real process management in `core/`. No action needed now, just FYI — we may want to adopt or adapt this protocol as the kernel-level interface later.

---

### 6. 🟡 `factory/_system.py` — dict placeholder pattern

The two-phase boot uses a raw dict `{"_class": ..., "agent_registry": ...}` instead of a proper instance or factory. Other services use actual instances or `None`. Consider using a proper placeholder class or lazy factory pattern for consistency.

---

### 7. 🟡 `bricks/llm/provider.py` — capability check removed

The change from `litellm.supports_function_calling(model=...)` to hardcoded `True` removes capability detection. This could break models that don't support function calling. If this is intentional for agent engine, please add a comment explaining why, or scope it to agent-engine callers only.

---

Let me know if you'd like help with any of these! Happy to pair on the kernel-related items.

PR #2815, opened 2026-03-07
feat: add PathLocalBackend as default (no CAS overhead)

joezhoujinjing:
## Summary
- Adds `PathLocalBackend` — stores files at actual filesystem paths (no CAS hashing/dedup overhead)
- Changes default backend from `"local"` (CAS) to `"path_local"` for simpler dev/minimal profiles
- Fixes kernel codepaths (`sys_unlink`, `write_batch`, `sys_rmdir`, `sys_rename`) to pass `backend_path` in context for path-based backends

## Details

`PathLocalBackend` uses the same local filesystem transport as `LocalBackend` but with direct path mapping instead of content-addressable storage. Files live at their actual virtual path under `root_path/files/` instead of being hash-sharded under `root_path/cas/`.

CAS remains available by setting `backend = "local"` in config (opt-in).

### Files changed
- **New**: `src/nexus/backends/path_local.py` — PathLocalBackend implementation with `@register_connector`
- **New**: `tests/unit/backends/test_path_local.py` — 15 tests (roundtrip, actual-path, overwrite, delete, dirs, registry, capabilities)
- `src/nexus/config.py` — default backend → `"path_local"`, added to allowed list
- `src/nexus/__init__.py` — route `path_local` vs `local` backend creation
- `src/nexus/core/nexus_fs.py` — add `backend_path` context to `sys_unlink`, `write_batch`, `sys_rmdir` recursive, skip `rename_file` for directories
- `src/nexus/backends/__init__.py`, `src/nexus/sdk/__init__.py` — register + export
- `tests/conftest.py`, `tests/unit/core/test_minimal_boot_mode.py` — use PathLocalBackend as default

## Test plan
- [x] 15 new PathLocalBackend unit tests pass
- [x] 39 minimal boot mode tests pass
- [x] 12 rename tests pass
- [x] 2218 core tests pass (0 failures)
- [x] 836 backend tests pass
- [ ] CI green

### DIFF
--- a/examples/tutorials/mount-management/test_path_local.sh
+++ b/examples/tutorials/mount-management/test_path_local.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# Test PathLocalBackend — files stored at actual paths, no CAS overhead.
+set -euo pipefail
+
+DIR=/tmp/test-path-local-$$
+trap 'rm -rf "$DIR"' EXIT
+
+echo "=== 1. Init workspace ==="
+uv run nexus init "$DIR"
+export NEXUS_DATA_DIR="$DIR/nexus-data"
+
+echo ""
+echo "=== 2. Write files ==="
+uv run nexus write /hello.txt "hello world"
+uv run nexus write /docs/readme.md "# README"
+
+echo ""
+echo "=== 3. Verify files at actual paths (not CAS-sharded) ==="
+test -f "$NEXUS_DATA_DIR/files/hello.txt" && echo "OK: hello.txt at actual path"
+test -f "$NEXUS_DATA_DIR/files/docs/readme.md" && echo "OK: docs/readme.md at actual path"
+! test -d "$NEXUS_DATA_DIR/cas" && echo "OK: no cas/ directory"
+
+echo ""
+echo "=== 4. Read back ==="
+OUT=$(uv run nexus cat /hello.txt)
+[ "$OUT" = "hello world" ] && echo "OK: read matches"
+
+echo ""
+echo "=== 5. Delete ==="
+echo "y" | uv run nexus rm /hello.txt
+! test -f "$NEXUS_DATA_DIR/files/hello.txt" && echo "OK: file removed from disk"
+
+echo ""
+echo "=== 6. Rename ==="
+echo "y" | uv run nexus move /docs/readme.md /docs/info.md
+test -f "$NEXUS_DATA_DIR/files/docs/info.md" && echo "OK: file moved on disk"
+! test -f "$NEXUS_DATA_DIR/files/docs/readme.md" && echo "OK: old path gone"
+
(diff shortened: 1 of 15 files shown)


joezhoujinjing (commented):
Review based on `docs/architecture/backend-architecture.md`:

1. **Architecture Violation in `src/nexus/config.py`:**
Changing the global default backend in `src/nexus/config.py` from `"local"` (CAS) to `"path_local"` violates the architecture guidelines in `docs/architecture/backend-architecture.md` (Section 2.2). The design explicitly states that CAS Addressing is the _"Default for all Nexus-owned storage, snapshots, versioning, federation replication"_. Changing this global default will break deduplication, zero-copy COW snapshots, and federation replication by default. If you want `path_local` for minimal/dev profiles, you should override `backend` only within those specific profile initialization paths or config files (e.g. `DeploymentProfile`), keeping the global `NexusConfig` default as `"local"`.

2. **Bug in `src/nexus/__init__.py` backend instantiation:**
In `connect()`, the logic was changed to:
```python
if cfg.backend == "local":
    backend = LocalBackend(root_path=Path(data_dir).resolve())
else:
    from nexus.backends.path_local import PathLocalBackend
    backend = PathLocalBackend(root_path=Path(data_dir).resolve())
```
This logic hardcodes the fallback to `PathLocalBackend` for **any** backend string that isn't `"local"`. This means if a user sets `cfg.backend = "gcs"` or another valid backend, it will silently ignore their choice and instantiate a `PathLocalBackend`. You should rely on `ConnectorRegistry` or a proper factory pattern to instantiate the backend dynamically, or at least use `elif cfg.backend == "path_local":` and raise an error or fall back to the registry for other types.

3. **Core Kernel `OperationContext` Changes:**
Adding `backend_path` to `OperationContext` in `write_batch`, `sys_unlink`, and `sys_rmdir` for `path_local` support is conceptually correct for path-based external backends, but ensuring that CAS remains the first-class citizen is critical.

Please revert the global default change in `src/nexus/config.py` and fix the backend instantiation logic in `src/nexus/__init__.py` to support other backend types properly.

joezhoujinjing (commented):
Looks much better!

1. **Restored Global Default**: Thanks for reverting the default in `src/nexus/config.py` to `"local"`. This preserves CAS behavior (deduplication, snapshots, etc.) as the baseline for Nexus-owned storage.
2. **Improved Instantiation**: The logic in `src/nexus/__init__.py` now correctly identifies `path_local` explicitly before falling back to the standard `LocalBackend`.
3. **Tests**: Verified that the test suite still defaults to `PathLocalBackend` in `tests/conftest.py` and `test_minimal_boot_mode.py`, which is appropriate for the faster test execution and simpler setup you intended.

The PR now aligns with the architectural design docs while still achieving the goal of providing a simpler, path-based local storage option.

PR #2822, opened 2026-03-07
feat: /etc/conf.d file-based config with VFS mount

joezhoujinjing:
## Summary
- Introduce Linux-style `/etc/conf.d/` brick configuration pattern
- New `nexus.etc` module — each brick calls `get_brick_config("name")` to load its own TOML config
- Auto-mount `$STATE_DIR/etc` as `/etc` in VFS (readonly) — config self-hosting
- `NEXUS_STATE_DIR` env var (default `~/.nexus`) as the single bootstrap parameter
- `nexus config init` command to generate default conf.d files
- Ship default `etc/conf.d/` files in repo (all commented out, self-documenting)
- POC: mount brick reads `auto_sync` from `/etc/conf.d/mounts` with env var fallback

## Design

Boot sequence:
1. `NEXUS_STATE_DIR` → hardcode kernel paths (metastore, data, etc)
2. Boot kernel — zero config needed
3. Mount `$STATE_DIR/etc` → `/etc` (readonly PathLocalBackend)
4. Bricks call `get_brick_config()` to load their own `/etc/conf.d/{name}`

Each brick loads its own config (like Linux daemons), no factory coupling.
Precedence: CLI args > env vars (deprecated) > conf.d files > profile defaults.

## Test plan
- [x] 16 unit tests for `nexus.etc` (TOML loading, missing files, malformed TOML, env var resolution)
- [x] 5 unit tests for `/etc` VFS mount (readonly, routing, skip when missing, conf.d integration)
- [x] 130 deployment profile/boot tests still pass
- [x] All pre-commit hooks pass (ruff, mypy, brick imports check)


windoliver:
1. installed CLI builds will not have any default conf.d files, so both nexus config init and first-boot seeding break outside a source checkout.
     config_cmd.py line 49:49 and __init__.py line 540:540 assume repo-level etc/conf.d, but packaging only discovers src packages in pyproject.toml line 316:316. I
     built the wheel locally; it contained no etc/conf.d entries, _find_confd_source() returned None, and _seed_etc_defaults() created nothing.
  2. /etc is described as read-only, but it is mounted writable. __init__.py line 495:495 says “readonly”, while __init__.py line 550:550 calls add_mount("/
     etc", etc_backend) without readonly=True. The router default is writable in router.py line 91:91, and write blocking only happens when that flag is set in
     router.py line 177:177. In standalone mode permissions are permissive by default in __init__.py line 369:369, so config can be mutated through the VFS.
  3. bootstrapping /etc is not metadata-only; it rewrites every host file on startup. __init__.py line 553:553 says the backend write is “effectively a no-
     op”, but __init__.py line 561:561 calls sys_write(), which performs a real backend write in nexus_fs.py line 2054:2054, increments version in nexus_fs.py line
     2061:2061, and emits a write event in nexus_fs.py line 2087:2087. For PathLocalBackend, that becomes an actual file replace in path_backend.py line 140:140 and
     local_transport.py line 77:77.

 Directionally, a nexus config command and file-backed brick config make sense for the CLI. I would not merge this version yet; it needs packaged defaults plus a
  clear /etc contract: either truly readonly with metadata bootstrap, or intentionally writable with the docs/comments changed and startup no longer rewriting
  files.

PR #3037, opened 2026-03-15
fix: tolerate chown failure on macOS Docker bind mounts

windoliver:
## Summary
- `nexus up` crash-loops on macOS because the entrypoint runs `chown -R nexus:nexus /app/data` which fails on VirtioFS bind mounts
- macOS Docker Desktop does not support `chown` on bind-mounted host directories — it returns "Permission denied"
- The container restarts endlessly, timing out the health check after 124s
- Fix: `chown ... 2>/dev/null || true` — the ownership change is best-effort; on macOS, Docker maps access through the host uid regardless

## Root cause
```
nexus-1  | chown: changing ownership of '/app/data/cas': Permission denied
nexus-1  | chown: changing ownership of '/app/data/tls/ca-key.pem': Permission denied
...  (repeats on every restart)
```

`set -e` in the entrypoint kills the script on the chown failure before the server starts.

## Test plan
- [ ] `nexus init --preset demo && nexus up` on macOS (Apple Silicon) — server starts
- [ ] `nexus up` on Linux — chown succeeds as before, no behavior change


elfenlieds7:
⚠️ **This PR was auto-closed** due to a git history rewrite on 2026-03-16 (contributor identity unification via `git filter-repo`).

**Replacement PR: #3055**

If you were working on this branch, your local clone has stale history. Run:
```
git fetch --all && git reset --hard origin/fix/entrypoint-chown-macos
```
Then push new commits to the replacement PR #3055.

PR #3043, opened 2026-03-16
fix(#3030): Rust correctness and performance fixes across 4 crates

### LINKED ISSUE #3030: Rust: High-priority correctness and performance issues
## Summary

This issue has been reduced to the problems that are still supported by the current tree. Removed items were either already fixed, overstated, or too speculative to keep as high-priority bugs.

---

### 1. [nexus_raft] `apply()` advances `last_applied` even if command execution fails

**File:** `rust/nexus_raft/src/raft/state_machine.rs:988-994`

`self.execute(command)` is called first, but `last_applied` is updated and persisted unconditionally afterward. If execution returns a storage error, the entry is still marked applied and may be skipped after restart.

### 2. [nexus_raft] `advance()` applies committed entries before persisting ready entries

**File:** `rust/nexus_raft/src/raft/node.rs:782-816`

Committed entries are applied to the state machine before `ready.entries()` are appended to storage. A crash in that window can leave the application state ahead of the durable Raft log.

### 3. [nexus_raft] Snapshot restore is not atomic

**File:** `rust/nexus_raft/src/raft/state_machine.rs:1028-1048`

`restore_snapshot()` clears metadata and locks, then repopulates them key-by-key, then persists `last_applied`. A crash in the middle can leave a partially restored snapshot on disk.

### 4. [nexus-fuse] Partial write path silently drops existing content when the read step fails

**File:** `nexus-fuse/src/fs.rs:556-577`

For `offset != 0`, the implementation does read-modify-write. If `read_cached()` fails, it falls back to `Vec::new()` and proceeds, which can zero or truncate the file prefix instead of returning `EIO`.

### 5. [nexus-fuse] Not-found handling depends on string matching

**Files:** `nexus-fuse/src/fs.rs:252`, `nexus-fuse/src/fs.rs:446`, `nexus-fuse/src/fs.rs:518`, `nexus-fuse/src/client.rs:169-176`

The client and FUSE layer classify errors by checking whether the message contains `"not found"`. This is fragile and can misclassify responses if server wording changes.

---

### 6. [nexus_pyo3] `read_files_bulk` defeats mmap by copying file contents into `Vec<u8>`

**File:** `rust/nexus_pyo3/src/io.rs:46-77`

Both the serial and parallel paths call `mmap.to_vec()`, so large files are fully copied into Rust-owned buffers before being copied again into Python objects.

### 7. [nexus_pyo3] SIMD wrappers copy full input arrays and do not release the GIL

**File:** `rust/nexus_pyo3/src/simd.rs`

The exported functions take owned `Vec<f32>` / `Vec<Vec<f32>>`, forcing full copies from Python. They also perform the work while the Python call still holds the GIL.

### 8. [nexus_core] `add_direct_subjects` does a full `tuple_index` scan

**File:** `rust/nexus_core/src/rebac/mod.rs:466-478`

This helper iterates every tuple to find direct subjects for one `(object, relation)` pair even though the graph already maintains secondary indexes.

---

## Priority

These remain high-priority because they can cause durable state divergence, skipped Raft entries, corrupted writes, or major overhead on common hot paths.

windoliver:
## Summary

Fixes 8 high-priority correctness and performance issues across `nexus_raft`, `nexus-fuse`, `nexus_pyo3`, and `nexus_core`. Includes 19 new tests.

### Raft safety (critical)
- **Persist-before-apply**: Reorder `advance()` to persist entries/hard state before applying committed entries (Raft paper §3, raft-rs README)
- **Panic on storage error**: `apply()` now panics on non-deterministic storage errors during committed-entry apply, matching etcd/CockroachDB behavior to prevent silent state divergence
- **Atomic snapshot restore**: `restore_snapshot()` now executes all clears + inserts in a single redb write transaction — crash mid-restore rolls back to previous state

### FUSE correctness
- **Partial write data corruption**: Return `EIO` when `read_cached()` fails during read-modify-write instead of silently zeroing the file prefix
- **Typed error classification**: Replace all 6 fragile `contains("not found")` string matches with typed `is_not_found()` method calls

### Performance
- **SIMD DRY + GIL release**: Extract generic `batch_cosine`/`top_k_cosine` helpers (267→195 lines), release GIL via `py.detach()` for batch operations
- **Zero-copy bulk reads**: Eliminate double-copy in `read_files_bulk` — collect `Mmap` objects in parallel, create `PyBytes` directly from mmap inside GIL
- **ReBAC O(1) lookup**: Replace O(n) `tuple_index` scan in `add_direct_subjects` with existing O(1) `reverse_adjacency` index

### New tests (+19)
- 3 Raft apply tests (idempotency guard, sequential advancement, gap handling)
- 4 snapshot restore tests (corrupt data preservation, empty data, overwrite, atomicity persistence)
- 7 SIMD pure-Rust tests (batch cosine, top-k ordering, i8 variants, validation)
- 5 FUSE mockito tests (RPC error classification, `is_not_found()` behavior)

## Test plan
- [x] `cargo test -p nexus_core` — 90 passed (includes ReBAC index fix)
- [x] `cargo test -p nexus_raft` — 42 passed (includes 7 new tests)
- [x] `cargo check -p nexus_pyo3` — compiles clean (tests require Python dev headers)
- [ ] `cargo test` for nexus-fuse (requires CI workspace resolution)
- [ ] Manual smoke test of FUSE mount + partial writes
- [ ] Verify SIMD results match expected values in Python integration tests

Closes #3030


elfenlieds7:
⚠️ **This PR was auto-closed** due to a git history rewrite on 2026-03-16 (contributor identity unification via `git filter-repo`).

**Replacement PR: #3052**

If you were working on this branch, your local clone has stale history. Run:
```
git fetch --all && git reset --hard origin/fix/3030-rust-correctness-perf
```
Then push new commits to the replacement PR #3052.

PR #3055, opened 2026-03-16
fix: tolerate chown failure on macOS Docker bind mounts

elfenlieds7:
Reopened after git history rewrite (email cleanup). Original PR: #3037


elfenlieds7:
Closing: the chown tolerance fix is already in develop (landed via another PR).

PR #3096, opened 2026-03-16
fix(auth): fall through to static API key when provider rejects token

windoliver:
## Summary

`resolve_auth()` in HTTP REST endpoints checked the auth provider first and returned immediately on `AuthResult(authenticated=False)`, **never reaching the static API key fallback**. This caused 401 errors on `/api/v2/aspects/` and `/api/v2/catalog/` endpoints when using `NEXUS_API_KEY` with database auth enabled.

The gRPC servicer already checks the static key first (works correctly). This aligns the HTTP REST path to match.

### Root cause

| Path | Check order | Behavior |
|------|-------------|----------|
| **gRPC** (working) | Static key first → provider | Static key matches → 200 |
| **HTTP REST** (broken) | Provider first → ~~static key~~ | Provider returns `authenticated=False` → **returns 401, never checks static key** |

The `DatabaseAPIKeyAuth` provider returns `AuthResult(authenticated=False)` (not `None`) for tokens not in its database. The code treated any non-None result as final. It also **cached the failure for 15 minutes**.

### Changes

1. **Provider rejection falls through**: When `result is None or not result.authenticated`, continue to static API key check instead of returning
2. **Singleflight fallthrough**: Coalesced waiters also fall through on provider rejection
3. **Cache guard**: Don't return cached `authenticated=False` results as cache hits

## Test plan
- [ ] `nexus aspects list /workspace/demo/restricted/internal.md` returns data (was: 401)
- [ ] `nexus catalog schema /workspace/demo/data/sales.csv` returns schema (was: 401)
- [ ] gRPC commands (cat, ls, search, versions) still work
- [ ] Auth with valid database-backed key still works (provider succeeds → cached → returned)
- [ ] Auth with invalid key still returns 401 (both provider and static key reject)

### DIFF
--- a/src/nexus/server/dependencies.py
+++ b/src/nexus/server/dependencies.py
@@ -195,7 +195,7 @@ async def resolve_auth(
         # Check cache first (15 min TTL) via CacheStoreABC
         _auth_cache: CacheStoreABC | None = getattr(_state, "auth_cache_store", None)
         cached_result = await _get_cached_auth(_auth_cache, token)
-        if cached_result:
+        if cached_result and cached_result.get("authenticated"):
             # Update per-request fields (zone header, agent ID, timing)
             # Safe: cached_result is already a copy from _get_cached_auth
             if x_nexus_zone_id:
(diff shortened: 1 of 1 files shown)


windoliver:
Superseded by combined PR #3100

PR #3119, opened 2026-03-17
fix(server,tui): snapshot service lookup + tab bar visibility

windoliver:
## Summary

- **snapshots.py**: Fix "Snapshot service not available" error on the versions panel. The lifespan wires the `TransactionalSnapshotService` to `app.state.transactional_snapshot_service`, but the REST router was only checking `nexus_fs._snapshot_service` (which is None). Now checks `app.state` first, then falls back.
- **tab-bar.tsx**: Brighten inactive tab labels from `#555555`/`#777777` to `#808080`/`#aaaaaa` so panel names like "Versions" and "Workflows" are visible on dark backgrounds.

## Test plan
- [ ] Versions panel (press `2`) should no longer show "Snapshot service not available"
- [ ] All tab names should be readable in the tab bar

### DIFF
--- a/packages/nexus-tui/src/app.tsx
+++ b/packages/nexus-tui/src/app.tsx
@@ -37,15 +37,15 @@ const ApiConsolePanel = lazy(() => import("./panels/api-console/api-console-pane
 
 const TABS: readonly Tab[] = [
   { id: "files", label: "Files", shortcut: "1" },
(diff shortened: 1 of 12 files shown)


windoliver:
Superseded by #3126 which includes all these fixes plus editor keybinding, transaction tracking, and rollback fixes.

PR #3358, opened 2026-03-26
fix: prevent write-back pipe wakeup test hot spin

windoliver:
## Summary
- replace the immediate-return AsyncMock in the write-back pipe wakeup test with an async stub that yields
- keep the assertion focused on whether wait_for_signal is reached without recording an unbounded call history
- prevent the poll loop test from hot-spinning and exhausting memory

## Testing
- uv run pytest -q -o addopts='' tests/unit/services/test_write_back_pipe_wakeup.py


windoliver:
Superseded by the fix being applied directly to PR #3356's branch.

PR #3494, opened 2026-03-29
refactor: move metadata.put from kernel to CAS driver (ext4 pattern)

elfenlieds7:
## Summary
CAS engine now handles metadata.put() inside write_content(), like Linux ext4
updating inodes in write_iter() while VFS holds i_rwsem. Kernel just holds
VFS lock and calls the driver — doesn't know about CAS metadata.

## Design (Linux ext4 analogy)
```
Kernel: _vfs_locked(path, "write") → backend.write_content(content, context) → unlock → events
CAS:    write blob → _commit_metadata(hash, context) → metastore.put()  [inside driver]
PAS:    write file → return  [no metadata touch]
```

## Changes (5 files)
- `backends/base/cas_addressing_engine.py`: Inject metastore, `_commit_metadata()` after blob store
- `contracts/types.py`: Add `existing_metadata` to OperationContext
- `core/nexus_fs.py`: Remove metadata.put from standard CAS path, pass `_meta` to avoid redundant get
- `backends/storage/cas_local.py`: Wire metastore to engine via `set_metastore()`
- `factory/orchestrator.py`: Activate metadata down-sink at boot

## Test plan
- [ ] All pre-commit hooks pass
- [ ] CI green — backward compat fallback for tests without DI wiring

### DIFF
--- a/src/nexus/backends/base/cas_addressing_engine.py
+++ b/src/nexus/backends/base/cas_addressing_engine.py
@@ -116,6 +116,7 @@ class CASAddressingEngine(Backend):
         meta_cache: Any | None = None,
         on_write_callback: Any | None = None,
         cdc_engine: "ChunkingStrategy | None" = None,
+        metastore: Any | None = None,
     ) -> None:
         self._transport = transport
         self._backend_name = backend_name or f"cas-{transport.transport_name}"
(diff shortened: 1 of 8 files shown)


elfenlieds7:
Closing: iterative patches created a messy state. Will redo with cleaner approach — auto-wire metastore via DLC.mount() from the start, avoid kernel fallback entirely.

PR #3620, opened 2026-04-03
fix: guard nexus_fast imports so nexus-fs works without Rust extension

windoliver:
## Summary

`nexus-fs` (slim package) ships `nexus/core/**` which has bare `from nexus_fast import` at module level. Since `nexus-fs` doesn't depend on `nexus-fast` (Rust extension), these imports crash with `ImportError` at runtime.

**Fix:** Add `nexus/_rust_compat.py` — a centralised shim that imports all `nexus_fast` symbols with graceful `None` fallback. Replace all bare imports in `nexus/core/` modules with imports from the shim.

### Files changed

| File | Change |
|------|--------|
| `_rust_compat.py` | New — central shim, re-exports all nexus_fast symbols |
| `metastore.py` | `RustDCache` via shim + None guard |
| `path_utils.py` | 9 path util imports via shim |
| `mount_table.py` | `RustPathRouter`, `canonicalize_path`, `extract_zone_id` via shim |
| `kernel_dispatch.py` | `HookRegistry`, `ObserverRegistry`, `PathTrie` via shim |
| `hash_fast.py` | `hash_content_py`, `hash_content_smart_py` via shim |
| `pipe.py`, `stream.py`, `shm_pipe.py`, `shm_stream.py` | Buffer classes via shim |

## Test plan

- [x] `from nexus._rust_compat import RUST_AVAILABLE` returns `True` with Rust, `False` without
- [x] All `nexus.core.*` modules import without error when Rust is absent
- [x] All existing functionality preserved when Rust is present

### DIFF
--- a/src/nexus/_rust_compat.py
+++ b/src/nexus/_rust_compat.py
@@ -1,0 +1,211 @@
+"""Rust extension compatibility shim.
+
+Centralises all ``nexus_fast`` imports so that the slim ``nexus-fs``
+package (which does not depend on the Rust extension) can still import
+``nexus.core.*`` modules without crashing at import time.
+
+Every symbol is re-exported as its original name.  When ``nexus_fast``
+is unavailable the symbol is set to ``None`` and call-sites must
+guard with ``if _Symbol is not None:`` before use.
+
+Usage in other modules::
+
+    from nexus._rust_compat import RustDCache  # None when Rust unavailable
+
+This avoids scattering try/except blocks across dozens of files.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+# -- Attempt bulk import of the Rust extension --------------------------------
+
+try:
+    import nexus_fast as _nf
+
+    RUST_AVAILABLE: bool = True
+except ModuleNotFoundError:
+    # Module genuinely not installed — slim/nexus-fs mode.
+    _nf = None  # type: ignore[assignment]
+    RUST_AVAILABLE = False
+except Exception:
+    # Module is installed but broken (ImportError, NameError, ABI skew,
+    # missing transitive dep, init-time crash). Catch broad Exception to
+    # prevent ANY import-time failure from escaping the shim.
+    _nf = None  # type: ignore[assignment]
+    RUST_AVAILABLE = False
+
+# Whether the import found a module at all (even if broken).
+# ModuleNotFoundError = not installed. Other ImportError = installed but broken.
+# hash_fast uses this to decide: "no extension" (SHA-256 ok for pure-Python
+# environments) vs "broken extension" (SHA-256 never ok — fail closed).
+RUST_EXTENSION_INSTALLED: bool = True  # assume installed unless ModuleNotFoundError
+try:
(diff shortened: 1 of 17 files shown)


elfenlieds7 (commented):
## Kernel Architecture Review

### Problem: Valid ✓
`nexus-fs` without Rust ext crashes on bare `from nexus_fast import`. Centralized shim is the right idea.

### Conflict: `_rust_compat.py` exports 6 symbols that no longer exist

PR #3621 (Kernel boundary collapse, pending on this branch) **deletes** these types from `nexus_fast`:

| Symbol in shim | Status after #3621 |
|---|---|
| `RustDCache` | ❌ Deleted — dcache is Kernel-internal |
| `RustPathRouter` | ❌ Deleted — router is Kernel-internal |
| `PathTrie` | ❌ Deleted — trie is Kernel-internal |
| `HookRegistry` | ❌ Deleted — hooks are Kernel-internal |
| `ObserverRegistry` | ❌ Deleted — observers are Kernel-internal |
| `SyscallEngine` | ❌ Renamed to `Kernel` (PR 2), constructor changed to 0-arg (PR 3) |

After #3621, `nexus/core/` modules **no longer import** any of these 6 symbols. They all go through `Kernel` proxy methods:
- `metastore.py`: `self._kernel.dcache_put(...)` (not `RustDCache()`)
- `mount_table.py`: `self._kernel.add_mount(...)` (not `RustPathRouter()`)
- `kernel_dispatch.py`: `self._kernel.trie_register(...)`, `self._kernel.register_hook(...)` (not `PathTrie()`, `HookRegistry()`)
- `nexus_fs.py`: `Kernel()` with 0 args (not `SyscallEngine(dcache, router, trie, lock)`)

### Recommended path

**Option A (preferred): Merge #3621 first → rebase this PR**

After rebase, `_rust_compat.py` core group simplifies dramatically:
```python
_CAPABILITY_GROUPS = {
    "core": ("Kernel", "RustRouteResult", "SysReadResult", "SysWriteResult"),
    # hash, ipc, storage groups stay as-is
}
```

And the `nexus_fs.py` changes reduce to just:
```python
from nexus._rust_compat import Kernel as _Kernel
# ...
self._kernel = _Kernel() if _Kernel is not None else None
```

**Option B: Merge this PR now → clean up after #3621**

If nexus-fs is blocking and can't wait: merge as-is, accept that 6 shim exports will become dead code when #3621 lands, and clean up in a follow-up PR. The shim won't break — it just exports None for symbols that no longer exist in Rust.

### Other observations (non-blocking)

1. **`hash_fast.py` fail-closed guard**: Good defensive design — `RUST_EXTENSION_INSTALLED` preventing silent SHA-256 fallback when Rust is stale.

2. **Capability groups**: Clean separation. After #3621 the core group shrinks but hash/ipc/storage stay unchanged.

3. **`nexus_fs.py` changes**: Still reference `SyscallEngine` which was renamed to `Kernel` in PR 2 (already merged to this branch). This will fail at runtime on current develop. Needs update regardless of which option you choose.

PR #3730, opened 2026-04-10
feat(#3701): stateless files=[...] filter + unified HTTP grep/glob surface

### LINKED ISSUE #3701: Stateful search result sets with `--from` chaining
## Problem

Nexus search is **stateless** — every search/grep call returns results independently. There is no way for an agent to say "grep within the files I already found." This forces agents into a pattern where each search resets their working set, preventing deep exploration:

```
# Today: stateless — each call is independent
search "authentication" → 10 results
search "JWT" → 10 DIFFERENT results, no connection to the first search
# Agent can never narrow within a previous result set
```

This is the single biggest bottleneck for agent search quality. Agents that can anchor a result set and iteratively narrow within it consistently find deeper results than agents doing independent searches.

## Proposal

Search and grep return a **result set ID**. Subsequent calls accept `--from <id>` to operate only within that set.

```
# Stateful: anchor → narrow → drill
search "authentication" → r_abc123 (10 files)
grep "JWT" --from r_abc123 → r_def456 (4 files within those 10)
grep "refresh.*token" --from r_def456 → r_ghi789 (2 files)
search "token rotation" --from r_abc123 → scoped semantic search within the 10
```

### New API Surface

```
# Search/grep already exist — they just gain result_id in response + --from in request
GET  /api/v2/search/query?q=auth                          → { result_id: "r_abc", results: [...] }
GET  /api/v2/search/query?q=JWT&from=r_abc                → scoped search within r_abc
POST /api/v2/search/grep?pattern=token&from=r_abc          → grep within r_abc's files

# Tag accumulation — merge multiple searches into one set
GET  /api/v2/search/query?q=auth&tag=security              → r_abc (15 files, tagged "security")
GET  /api/v2/search/query?q=RBAC&tag=security              → r_abc (25 files, merged into same tag)
GET  /api/v2/search/query?q=OAuth&tag=security              → r_abc (33 unique files)
POST /api/v2/search/grep?pattern=vuln&from=tag:security     → grep across all 33

# Result set management
GET    /api/v2/search/results                              → list saved result sets
GET    /api/v2/search/results/:id                          → list files in a set
DELETE /api/v2/search/results/:id                          → manual cleanup
```

### MCP Tools

- `nexus_semantic_search` → returns `result_id` in response
- `nexus_grep` → accepts optional `from_result` param, returns `result_id`
- `nexus_glob` → same pattern

### CLI

- `nexus search query "auth"` → prints result ID
- `nexus search grep "JWT" --from r_abc123` → grep within set
- Pipe support: `nexus search query "auth" | nexus search grep "JWT"` (result ID on stdout)

### Storage

Simple table or in-memory store with TTL:

```sql
CREATE TABLE search_result_sets (
    id          VARCHAR(36) PRIMARY KEY,
    zone_id     VARCHAR(255) NOT NULL,
    tag         VARCHAR(255),          -- optional named tag for accumulation
    file_paths  JSONB NOT NULL,        -- set of file paths in this result
    created_by  VARCHAR(255),
    created_at  TIMESTAMP DEFAULT NOW(),
    expires_at  TIMESTAMP              -- auto-cleanup via TTL
);
CREATE INDEX idx_result_sets_zone ON search_result_sets (zone_id);
CREATE INDEX idx_result_sets_tag ON search_result_sets (zone_id, tag);
```

Alternative: in-memory dict in the daemon with configurable TTL (simpler, lost on restart, fine for ephemeral session state).

### Behavior

- `--from` filters the operation to only files in the referenced result set
- For `search --from`: semantic/hybrid search runs normally but results are intersected with the set
- For `grep --from`: only files in the set are scanned (trigram index candidates ∩ result set)
- Tag accumulation: `UPSERT` — merge new file paths into existing tag's set (deduped)
- TTL: default 1 hour, configurable via `NEXUS_RESULT_SET_TTL`
- No `--from` = current behavior (backward compatible)

### Files to Touch

- `src/nexus/bricks/search/daemon.py` — result set storage + `--from` filtering in search/grep paths
- `src/nexus/server/api/v2/routers/search.py` — `from` and `tag` query params, result set management endpoints
- `src/nexus/bricks/mcp/server.py` — `from_result` param on MCP tools, return `result_id`
- `src/nexus/cli/commands/search.py` — `--from` flag, result ID output
- `alembic/versions/` — migration for `search_result_sets` table (if DB-backed)

windoliver:
Closes #3701.

## Summary

- Adds stateless `files=[...]` narrowing to `grep`/`glob` across SearchService, HTTP, and MCP — agents can now pass a pre-narrowed working set to drill down without a server-side state subsystem.
- First-time `GET /api/v2/search/grep` and `GET /api/v2/search/glob` HTTP endpoints, wired through SearchService with full ReBAC enforcement (previously these only existed as MCP tools and bypassed the router-level permission filter).
- Fixes a pre-existing pagination bug in MCP `nexus_grep` that silently capped results at 100, replaces the `fetch_limit = limit * 3` magic number with a named constant + `permission_denial_rate`/`truncated_by_permissions` instrumentation, and collapses duplicated response-dict builders into a shared serializer.

## Why this shape (pivot from the original #3701 proposal)

The issue originally proposed server-side stateful result-set IDs with TTL, tag accumulation, and a Postgres migration. An interactive review surfaced that:

1. Claude Code's own search pattern is **stateless by design** — agents keep the file list and pass it back as narrowing. Stateful IDs hide working-set visibility from the agent's own context and introduce TTL/storage/identity concerns.
2. The proposal assumed HTTP grep/glob already existed (`Search/grep already exist — they just gain result_id in response`). They did not. The review-level discovery forced a precursor "unify grep/glob" step before anything else could land.
3. The 9-row edge-case matrix (traversal, cross-zone, stale, size cap, dedupe, normalization, permission intersection, file_pattern interaction, empty list) is the same whether the working set lives server-side or in the agent's context — but client-side removes a subsystem's worth of TTL/expiry/sliding-refresh design.

So we pivoted to a stateless `files=[...]` parameter. If real usage shows the stateless approach is insufficient, the server-side result-set path can still be added later — the contract (in-memory daemon dict + LRU, `r_{12ch}` IDs, hard 410 on expiry, no sliding TTL) is already decided in the review log.

### Production code
- **`src/nexus/lib/pagination.py`** — adds `build_paginated_list_response`, the shared offset/limit envelope used by MCP `nexus_grep`/`nexus_glob` and the new HTTP endpoints. Lives in `lib/` (not `bricks/search/`) to stay cross-brick-safe — the MCP brick cannot import from the search brick per importlinter rules.
- **`src/nexus/server/api/v2/routers/search.py`**
  - New `GET /api/v2/search/grep` and `GET /api/v2/search/glob` endpoints, both enforcing `_apply_rebac_filter`.
  - `_serialize_search_result` collapses the 25-line dict-builder previously duplicated across the graph and non-graph branches of `search_query`.
  - `_REBAC_OVERFETCH_FACTOR` constant replaces the hardcoded `3` magic number.
  - `_rebac_denial_stats` adds `permission_denial_rate` and `truncated_by_permissions` fields to every response envelope + a server-side warning log when denial rate > 50% causes undercount.
- **`src/nexus/bricks/search/search_service.py`**
  - `grep` and `glob` accept `files: list[str] | None = None`.
  - `_validate_and_normalize_files` implements the 9-row edge-case spec: empty list short-circuit, traversal rejection, cross-zone rejection, dedupe, stale-silent-skip with `stale_count`, 10k size cap, `file_pattern` intersection, path normalization, and permission intersection via `self.list`.
  - `_select_grep_strategy` gains a threshold bypass: when `files=[...]` is set and the list is smaller than `FILES_FILTER_TRIGRAM_THRESHOLD` (200), trigram is bypassed in favour of direct scanning. When larger, trigram runs with over-fetch and post-filtering.
  - Tree-walk short-circuit: `self.list(path, recursive=True)` is only called inside the validator for the permission intersection when `files` is supplied — no redundant full-tree walk.
- **`src/nexus/bricks/mcp/server.py`**
  - Fixes the pre-existing `nexus_grep` pagination bug: now forwards `max_results=max(limit + offset, 1)` instead of relying on the default 100 cap, so `limit=200` actually returns 200 matches.
  - `nexus_grep` and `nexus_glob` gain a `files: list[str] | None = None` parameter.
  - Uses `build_paginated_list_response` from `nexus.lib.pagination` (not the bricks-search helper that a previous draft placed there — that would have violated the brick-independence importlinter contract).
  - Adds TODO comments documenting the known pre-existing MCP permission gap (`Context` carries only an API key, not subject identity) with a pointer to the follow-up issue.

### Tests
- **`tests/unit/server/routers/test_search_rebac_filter.py`** (new, 37 tests) — backfills coverage for `_apply_rebac_filter`, `_normalize_path`, `_serialize_search_result`, `_compute_rebac_fetch_limit`, and `_rebac_denial_stats`. These helpers previously had zero test coverage on the search permission boundary.
- **`tests/unit/lib/test_pagination.py`** (new, 12 tests) — covers the new paginated envelope builder + round-trip sanity for the existing cursor encoder.
- **`tests/integration/services/test_search_service.py`** — adds `TestGrepFilePattern` (8 tests backfilling coverage for the pre-existing `file_pattern` param), `TestValidateAndNormalizeFiles` (11 tests for the 9-row edge-case matrix), and `TestGrepFilesParam`/`TestGlobFilesParam` (10 tests for end-to-end short-circuit, intersection semantics, and size-cap rejection through grep/glob).
- **`tests/integration/server/api/v2/routers/test_search_http_grep_glob.py`** (new, 36 tests) — HTTP integration tests via FastAPI TestClient: happy paths, pagination, error paths (400/422/500/503), ReBAC interaction with `denial_rate`/`truncated` instrumentation, envelope parity between grep/glob, cross-endpoint parity for the same user + `files=[...]` parameter flow.
- **`tests/unit/bricks/mcp/test_mcp_server_tools.py`** — updates existing assertions to match the new `max_results=` and `files=` signatures, adds new tests for `files=[...]` forwarding and the 14A pagination fix.
- **`tests/benchmarks/test_search_benchmarks.py`** — 3 benchmarks measuring validator overhead at small (10) / threshold (200) / size-cap (10k) list sizes.

## Stats

- 10 files changed, +2488/-90
- ~130 targeted new tests, all green
- 298 tests passing across all touched test paths
- Pre-existing `test_txtai_backend.py` failures unrelated to this PR (reproduced on the `develop` baseline before any of my changes landed)

## Deferred

- **MCP permission unification** — a proper fix requires threading an API-key → `OperationContext` lookup through MCP tool handlers. Infrastructure doesn't exist yet. Scoped out per Direction X in the review; follow-up issue linked below. HTTP grep/glob go through proper ReBAC; MCP grep/glob retain their pre-existing behaviour with documented TODO comments.
- **Server-side result-set IDs with TTL** (original #3701 proposal) — parked. Can be added as phase 2 if real usage shows the stateless approach is insufficient. The contract (in-memory daemon dict with LRU, `r_{12ch}` IDs, hard 410 on expiry, no sliding TTL) is already decided in the review log.

## Test plan

- [x] `pytest tests/unit/server/routers/test_search_rebac_filter.py` — 37 passing
- [x] `pytest tests/unit/lib/test_pagination.py` — 12 passing
- [x] `pytest tests/integration/services/test_search_service.py::TestGrepFilePattern` — 8 passing
- [x] `pytest tests/integration/services/test_search_service.py::TestValidateAndNormalizeFiles` — 11 passing
- [x] `pytest tests/integration/services/test_search_service.py::TestGrepFilesParam tests/integration/services/test_search_service.py::TestGlobFilesParam` — 10 passing
- [x] `pytest tests/integration/server/api/v2/routers/test_search_http_grep_glob.py` — 36 passing
- [x] `pytest tests/unit/bricks/mcp/test_mcp_server_tools.py::TestSearchTools` — passing (including the updated assertions for the new signatures)
- [x] `pytest tests/integration/services/test_search_router.py` — existing suite still passing (refactor did not regress the `search_query` endpoint)
- [x] `ruff check` — clean across all 10 changed files
- [x] `ruff format --check` — clean
- [x] `mypy` — clean (no new `# type: ignore` suppressions)
- [x] Brick-independence importlinter — clean
- [ ] Manual verification of the new HTTP endpoints against a live daemon (worth doing before merge — the integration tests use mocked SearchService)
- [ ] Review whether `FILES_FILTER_TRIGRAM_THRESHOLD = 200` is tuned appropriately for your typical zone sizes (the constant has a benchmark-source comment — benchmarks are included but not calibrated against a realistic corpus)

### DIFF
--- a/rust/nexus_core/src/rebac/graph.rs
+++ b/rust/nexus_core/src/rebac/graph.rs
@@ -252,6 +252,7 @@ pub fn compute_permission_interned(
             InternedRelationConfig::TupleToUserset {
                 tupleset,
                 computed_userset,
+                skip_reverse,
             } => {
                 // tupleToUserset checks BOTH directions:
                 //
@@ -262,6 +263,12 @@ pub fn compute_permission_interned(
                 // Reverse (group pattern): others have tupleset relation ON object
                 //   group_viewer = {tupleset: "direct_viewer", computedUserset: "member"}
                 //   group:team → direct_viewer → file:/path, then check member on group:team
+                //
+                // Fix nexi-lab/nexus#3733 Bug A: the reverse direction is
+                // skipped for ``parent`` tuplesets because it inverts
+                // parent semantics (finds children instead of the parent)
+                // and grants permission based on owning any child —
+                // which is a privilege escalation.
                 let mut allowed = false;
 
                 // Forward: object as subject → find objects it points to
@@ -282,8 +289,9 @@ pub fn compute_permission_interned(
                     }
                 }
 
-                // Reverse: find subjects that have tupleset relation ON object
-                if !allowed {
+                // Reverse: find subjects that have tupleset relation ON object.
+                // Skipped for ``parent`` tupleset (see comment above).
+                if !allowed && !skip_reverse {
                     let reverse_targets = graph.find_subjects_for_object(object, *tupleset);
                     for target in &reverse_targets {
                         if compute_permission_interned(
--- a/rust/nexus_core/src/rebac/mod.rs
+++ b/rust/nexus_core/src/rebac/mod.rs
@@ -305,8 +305,23 @@ pub fn compute_permission(
                     }
                 }
 
-                // Reverse: find subjects that have tupleset relation ON object
-                if !allowed {
+                // Reverse: find subjects that have tupleset relation ON object.
+                //
+                // Fix nexi-lab/nexus#3733 Bug A: skip the reverse (group)
+                // pattern when the tupleset relation is "parent". For a
+                // parent relation, the forward pattern is the ONLY correct
+                // direction — "alice is parent_owner of Y iff alice owns
+                // parent(Y)". The reverse pattern finds Y's CHILDREN
+                // instead and incorrectly grants parent permission based
+                // on owning any child, causing a privilege escalation
+                // where owning /workspace/public grants access to all
+                // sibling files under /workspace/.
+                //
+                // The equivalent Python guards are in
+                // bricks/rebac/graph/bulk_evaluator.py,
+                // bricks/rebac/graph/traversal.py, and
+                // bricks/rebac/graph/zone_traversal.py.
+                if !allowed && tuple_to_userset.tupleset != "parent" {
                     let reverse_targets =
                         graph.find_subjects_for_object(object, &tuple_to_userset.tupleset);
                     for target in &reverse_targets {
--- a/rust/nexus_core/src/types.rs
+++ b/rust/nexus_core/src/types.rs
@@ -130,6 +130,16 @@ pub enum InternedRelationConfig {
     TupleToUserset {
         tupleset: Sym,
         computed_userset: Sym,
+        /// Skip the reverse (group-style) pattern 2 evaluation.
+        ///
+        /// Set to ``true`` at build time for tupleset relations where
+        /// the reverse direction has broken semantics — specifically
+        /// ``parent``, where "X is parent_owner of Y iff X owns
+        /// parent(Y)" requires the forward direction only. The reverse
+        /// direction would find Y's CHILDREN and grant permission based
+        /// on owning any child, which is a privilege escalation.
+        /// See nexi-lab/nexus#3733 Bug A.
+        skip_reverse: bool,
     },
 }
 
@@ -149,10 +159,18 @@ impl InternedNamespaceConfig {
                         union: union.iter().map(|s| interner.get_or_intern(s)).collect(),
                     },
                     RelationConfig::TupleToUserset { tuple_to_userset } => {
+                        // Fix nexi-lab/nexus#3733 Bug A: pre-compute
+                        // skip_reverse flag at build time. For ``parent``
+                        // tuplesets the reverse direction is a privilege
+                        // escalation (owning any child would grant parent
+                        // access). Comparing the raw string here is
+                        // cheaper than interning "parent" separately.
+                        let skip_reverse = tuple_to_userset.tupleset == "parent";
                         InternedRelationConfig::TupleToUserset {
                             tupleset: interner.get_or_intern(&tuple_to_userset.tupleset),
                             computed_userset: interner
                                 .get_or_intern(&tuple_to_userset.computed_userset),
+                            skip_reverse,
                         }
                     }
                 };
--- a/rust/nexus_kernel/src/rebac.rs
+++ b/rust/nexus_kernel/src/rebac.rs
@@ -165,6 +165,7 @@ fn compute_permission_interned_shared(
             InternedRelationConfig::TupleToUserset {
                 tupleset,
                 computed_userset,
+                skip_reverse,
             } => {
                 // Forward: object as subject → find objects it relates to
                 let forward = graph
@@ -184,6 +185,12 @@ fn compute_permission_interned_shared(
                     });
                 if forward {
                     true
+                } else if *skip_reverse {
+                    // Fix nexi-lab/nexus#3733 Bug A: skip the reverse pattern
+                    // for ``parent`` tuplesets. See graph.rs for the full
+                    // rationale — the reverse direction inverts parent
+                    // semantics and causes privilege escalation.
+                    false
                 } else {
                     // Reverse: find subjects that have tupleset relation ON object
                     // (H25: must match graph.rs which checks both directions)
--- a/scripts/generate_rpc_params.py
+++ b/scripts/generate_rpc_params.py
@@ -230,8 +230,23 @@ def _default_repr(default: Any) -> str:
 
 
 def _is_context_param(name: str, annotation_str: str) -> bool:
-    """Return True if this parameter represents an OperationContext."""
-    if annotation_str in CONTEXT_TYPES:
+    """Return True if this parameter represents an OperationContext.
+
+    Handles bare forms (``OperationContext``, ``OperationContext | None``)
+    AND single-quoted double-forward-reference forms
+    (``'OperationContext | None'``) that leak through when the source
+    has ``context: "OperationContext | None"`` with `from __future__
+    import annotations` — inspect.signature returns the annotation as
+    a string that still contains the literal inner quotes.
+    """
+    # Strip any outer single/double quote pair before comparing — this
+    # catches the double-forward-reference case.
+    stripped = annotation_str.strip()
+    if (stripped.startswith("'") and stripped.endswith("'")) or (
+        stripped.startswith('"') and stripped.endswith('"')
+    ):
+        stripped = stripped[1:-1]
+    if stripped in CONTEXT_TYPES or annotation_str in CONTEXT_TYPES:
         return True
     # Also exclude _context params (internal convention)
     return name.startswith("_") and "context" in name.lower()
--- a/scripts/generate_rpc_params.py
+++ b/scripts/generate_rpc_params.py
@@ -76,7 +76,11 @@ EXCLUDED_METHODS: set[str] = {
     "get_dynamic_viewer_config",
     "read_with_dynamic_viewer",
     "apply_dynamic_viewer_filter",
-    "semantic_search",
+    # Codex review of #3701 (finding #3): semantic_search REMAINS exposed
+    # via SearchService and is called positionally by RemoteServiceProxy
+    # (e.g. ``proxy.semantic_search("needle", path="/docs", limit=5)``).
+    # Excluding it from METHOD_PARAMS broke that positional binding —
+    # restored so RPCProxyBase._get_param_names() can resolve "query".
     "semantic_search_stats",
     "initialize_semantic_search",
     "wait_for_changes",
@@ -87,6 +91,16 @@ EXCLUDED_METHODS: set[str] = {
 
 # Modules to scan for @rpc_expose methods.  We inspect each class in each
 # module and collect all methods that have ``_rpc_exposed = True``.
+#
+# Codex review of #3701 (finding #3): the previous list referenced stale
+# ``nexus.system_services.*`` module paths that were moved during the
+# kernel/services refactor. Those modules silently ImportError'd, the
+# codegen logged a warning, and the resulting METHOD_PARAMS was missing
+# entries for ``register_workspace``, ``register_agent``, and several
+# other still-exposed RPCs. RemoteServiceProxy / RPCProxyBase rely on
+# METHOD_PARAMS to map positional arguments to keyword names, so any
+# missing entry silently misserializes positional remote calls. Fix by
+# pointing at the canonical post-refactor locations.
 MODULES_TO_SCAN: list[str] = [
     "nexus.core.nexus_fs",
     # Search (moved to bricks tier, Issue #1287)
@@ -95,14 +109,24 @@ MODULES_TO_SCAN: list[str] = [
     "nexus.bricks.share_link.share_link_service",
     "nexus.bricks.versioning.version_service",
     "nexus.bricks.mount.mount_service",
-    "nexus.services.oauth.oauth_service",
-    "nexus.bricks.memory.memory_service",
-    "nexus.system_services.workspace.workspace_rpc_service",
-    "nexus.system_services.lifecycle.user_provisioning",
-    "nexus.system_services.agents.agent_service",
+    # OAuth credential service (moved into bricks/auth/oauth/)
+    "nexus.bricks.auth.oauth.credential_service",
+    # post-refactor canonical locations (was nexus.system_services.*)
+    "nexus.services.workspace.workspace_rpc_service",
+    "nexus.services.agents.agent_rpc_service",
+    "nexus.services.lifecycle.user_provisioning",
+    "nexus.services.acp.acp_rpc_service",
+    "nexus.services.metadata_export",
     # Brick services with @rpc_expose
     "nexus.bricks.mcp.mcp_service",
     "nexus.bricks.rebac.rebac_service",
+    # Server-side RPC services (Issue #2986: post-refactor split)
+    "nexus.server.rpc.services.snapshots_rpc",
+    "nexus.server.rpc.services.pay_rpc",
+    "nexus.server.rpc.services.governance_rpc",
+    "nexus.server.rpc.services.federation_rpc",
+    "nexus.server.rpc.services.events_rpc",
+    "nexus.server.rpc.services.audit_rpc",
 ]
 
 # Types that signal "operation context" — parameters with these types are
--- a/src/nexus/bricks/mcp/server.py
+++ b/src/nexus/bricks/mcp/server.py
@@ -17,6 +17,7 @@ from fastmcp import Context, FastMCP
 from nexus.bricks.mcp.formatters import format_response
 from nexus.bricks.mcp.tool_utils import handle_tool_errors, tool_error
 from nexus.contracts.filesystem.filesystem_abc import NexusFilesystem
+from nexus.lib.pagination import build_paginated_list_response
 
 logger = logging.getLogger(__name__)
 
@@ -667,6 +668,7 @@ async def create_mcp_server(
         limit: int = 100,
         offset: int = 0,
         response_format: str = "json",
+        files: list[str] | None = None,
         ctx: Context | None = None,
     ) -> str:
         """Search files using glob pattern with pagination.
@@ -677,6 +679,9 @@ async def create_mcp_server(
             limit: Maximum number of results to return (default: 100)
             offset: Number of results to skip (default: 0)
             response_format: Output format - "json" or "markdown" (default: "json")
+            files: Optional stateless narrowing working set (#3701). When
+                provided, the glob pattern is matched against this list
+                instead of walking the tree under ``path``.
 
         Returns:
             Formatted string with paginated search results containing:
@@ -690,33 +695,38 @@ async def create_mcp_server(
         Example:
             To find all Python files: nexus_glob("**/*.py", "/workspace")
             With pagination: nexus_glob("**/*.py", "/workspace", limit=50, offset=0)
+            Narrowed: nexus_glob("**/*.py", files=["/src/a.py", "/src/b.py"])
         """
         nx_instance: Any = _get_nexus_instance(ctx)
         _search = nx_instance.service("search")
         if _search is None:
             raise ValueError("SearchService not available — glob requires the search brick")
-        all_matches = _search.glob(pattern, path)
+        # NOTE(#3701): SearchService.glob is not passed an OperationContext
+        # here because the MCP transport's Context only carries a per-request
+        # API key, not (subject_id, zone_id, is_admin). Permission filtering
+        # therefore currently relies on whatever the NexusFilesystem connection
+        # enforces at the service-lookup layer. A proper fix requires
+        # threading an API-key → identity lookup into MCP tool handlers; see
+        # the follow-up issue linked in the #3701 review summary.
+        all_matches = _search.glob(pattern, path, files=files)
         total = len(all_matches)
 
         # Apply pagination
         paginated_matches = all_matches[offset : offset + limit]
-        has_more = (offset + limit) < total
 
         # Issue #538: Log truncation when results exceed limit
-        if has_more or offset > 0:
+        if (offset + limit) < total or offset > 0:
             logger.info(
                 f"[GLOB] Truncated {total} -> {len(paginated_matches)} results "
                 f"(offset={offset}, limit={limit})"
             )
 
-        result = {
-            "total": total,
-            "count": len(paginated_matches),
-            "offset": offset,
-            "items": paginated_matches,
-            "has_more": has_more,
-            "next_offset": offset + limit if has_more else None,
-        }
+        result = build_paginated_list_response(
+            items=paginated_matches,
+            total=total,
+            offset=offset,
+            limit=limit,
+        )
 
         return format_response(result, response_format)
 
@@ -736,6 +746,7 @@ async def create_mcp_server(
         limit: int = 100,
         offset: int = 0,
         response_format: str = "json",
+        files: list[str] | None = None,
         ctx: Context | None = None,
     ) -> str:
         """Search file contents using regex pattern with pagination.
@@ -747,6 +758,10 @@ async def create_mcp_server(
             limit: Maximum number of results to return (default: 100)
             offset: Number of results to skip (default: 0)
             response_format: Output format - "json" or "markdown" (default: "json")
+            files: Optional stateless narrowing working set (#3701). When
+                provided, grep searches only these files instead of walking
+                the tree. Agents should pass the file list from a previous
+                search/grep to drill down into its results.
 
         Returns:
             Formatted string with paginated search results containing:
@@ -760,33 +775,41 @@ async def create_mcp_server(
         Example:
             To get first 50 matches: nexus_grep("TODO", "/workspace", limit=50)
             To get next 50 matches: nexus_grep("TODO", "/workspace", limit=50, offset=50)
+            Narrowed: nexus_grep("TODO", files=["/src/a.py", "/src/b.py"])
         """
         nx_instance: Any = _get_nexus_instance(ctx)
         _search = nx_instance.service("search")
         if _search is None:
             raise ValueError("SearchService not available — grep requires the search brick")
-        all_results = await _search.grep(pattern, path, ignore_case=ignore_case)
+        # NOTE(#3701): See glob handler above for the MCP permission caveat.
+        # Also (#3701 Issue 14): we pass max_results so SearchService returns
+        # enough matches for the caller's requested page rather than silently
+        # capping at its default of 100.
+        all_results = await _search.grep(
+            pattern,
+            path,
+            ignore_case=ignore_case,
+            max_results=max(limit + offset, 1),
+            files=files,
+        )
         total = len(all_results)
 
         # Apply pagination
         paginated_results = all_results[offset : offset + limit]
-        has_more = (offset + limit) < total
 
         # Issue #538: Log truncation when results exceed limit
-        if has_more or offset > 0:
+        if (offset + limit) < total or offset > 0:
             logger.info(
                 f"[GREP] Truncated {total} -> {len(paginated_results)} results "
                 f"(offset={offset}, limit={limit})"
             )
 
-        result = {
-            "total": total,
-            "count": len(paginated_results),
-            "offset": offset,
-            "items": paginated_results,
-            "has_more": has_more,
-            "next_offset": offset + limit if has_more else None,
-        }
+        result = build_paginated_list_response(
+            items=paginated_results,
+            total=total,
+            offset=offset,
+            limit=limit,
+        )
 
         return format_response(result, response_format)
 
--- a/src/nexus/bricks/mcp/server.py
+++ b/src/nexus/bricks/mcp/server.py
@@ -76,7 +76,7 @@ def reset_request_api_key(token: contextvars.Token[str | None]) -> None:
 
 
 def _resolve_mcp_operation_context(nx_instance: NexusFilesystem) -> Any:
-    """Build an explicit ``OperationContext`` for MCP search calls.
+    """Resolve an explicit ``OperationContext`` for MCP search calls.
 
     Codex review #3 finding #1: the previous MCP grep/glob handlers
     called ``SearchService`` without any ``OperationContext`` at all,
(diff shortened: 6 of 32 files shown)


windoliver:
## Live validation against \`nexus up --build\` — results + fixes

Ran the full Docker stack from this branch's source and exercised every new endpoint end-to-end. **Three bugs surfaced that the mock-based unit/integration tests did not catch.** All three are fixed in commit 25cae4512 on this branch.

### Bug 1 — Path traversal returned 500, should be 400

The HTTP handlers caught \`ValueError\` → 400 but not \`InvalidPathError\`. The real \`SearchService._validate_and_normalize_files\` raises \`InvalidPathError\` on a traversal segment, which leaked through to the generic \`Exception\` handler as 500.

**Fix:** Extended the client-error catch in \`search_grep\` and \`search_glob\` to include \`InvalidPathError\`. Added two regression tests (\`test_grep_files_traversal_invalid_path_error_returns_400\` and \`test_glob_invalid_path_error_returns_400\`) so the mock-based suite now asserts on the real exception type.

### Bug 2 — \`files=[...]\` kwarg silently dropped by the RPC layer

The most subtle bug. The HTTP endpoints worked correctly because they invoke SearchService directly, but the remote SDK and MCP tools go through the gRPC/REST-RPC path:

\`\`\`
client → parse_method_params(method, params_dict) → GrepParams(**params)
       → handle_grep(nexus_fs, params, context)
       → search.grep(pattern, **kwargs)  # explicit allowlist of kwargs
\`\`\`

**Two layers** were stripping \`files\`:

1. \`src/nexus/server/_rpc_params_generated.py\` — auto-generated dataclasses that validate and extract RPC params. The committed version didn't have \`files\` on \`GrepParams\`/\`GlobParams\` because the generator hadn't been re-run since my signature changes. Running the regenerator wanted to touch unrelated pre-existing drift (\`sys_access\` → \`access\` rename that the overrides file still references), so I hand-added \`files: list[str] | None = None\` to \`GrepParams\` and \`GlobParams\` only, keeping the PR scope clean.
2. \`src/nexus/server/rpc/handlers/filesystem.py::handle_grep\`/\`handle_glob\` — the server-side RPC handlers explicitly allowlist kwargs before forwarding to SearchService. Added \`if hasattr(params, "files") and params.files is not None\` blocks that preserve the empty-list short-circuit (\`files=[]\` must reach SearchService intact, not be coalesced to \`None\`).

**Before fix** (remote SDK \`svc.grep(pattern="TODO", files=["/a.py", "/b.py"])\`): returned 4 matches (the baseline — \`files\` was dropped).
**After fix**: returns exactly 2 matches.

### Bug 3 — none; #13A trigram-bypass path verified live

Seeded 600 files (above \`GREP_TRIGRAM_THRESHOLD=500\`) and measured:

| Scenario | Result | Latency |
|---|---|---|
| baseline (no files filter) | 200 matches | 694ms |
| \`files=[50]\` — small-list bypass path | 50 matches | **20ms** (35x faster) |
| \`files=[300]\` — trigram + post-filter | 100 matches | 157ms |

The 35x speedup on the small-list bypass confirms the short-circuit avoids the full tree walk as designed.

### ReBAC instrumentation — live-populated

\`permission_denial_rate\` and \`truncated_by_permissions\` appear in every HTTP response. With the \`shared\` preset running as admin the values are 0.0 (no denial), but \`permission_filter_ms\` is populated with real timings confirming the helper is invoked on every request. Unit tests cover the non-zero scenarios; live testing with a partial-permission user is deferred (unit-test coverage plus the plumbing confirmation is enough).

### Live test matrix — all passing

- ✅ \`GET /api/v2/search/health\` — returns healthy envelope
- ✅ \`GET /api/v2/search/query\` — refactor didn't regress, \`_serialize_search_result\` produces the expected shape, \`denial_rate\`/\`truncated\` populate
- ✅ \`GET /api/v2/search/grep\` — basic, path filter, \`ignore_case\`, pagination (limit+offset), case-sensitive vs insensitive, \`before_context\`/\`after_context\`
- ✅ \`GET /api/v2/search/grep?files=...&files=...\` — narrowing, single-file, stale silent-skip
- ✅ Grep error paths: invalid regex → 400, empty pattern → 422, negative offset → 422, path traversal → 400 (after fix), no auth → 401
- ✅ \`GET /api/v2/search/glob\` — \`*.py\`, \`**/*.py\`, \`**/*.md\`, path prefix, \`files=[...]\` intersection, pagination, traversal → 400
- ✅ Remote SDK \`svc.grep(files=[...])\` via gRPC — 6 scenarios: baseline, narrow-to-2, empty, stale, glob mixed-ext, size cap rejection (after RPC fix)
- ✅ Trigram bypass: 600-file corpus, 50-file bypass path, 300-file trigram+post-filter path — all correct
- ✅ Unit + integration suite: 84 tests passing after the fixes

### Deferred (documented, not shipped)

- **Live ReBAC-denial scenario** — requires setting up a user with partial permissions via \`nexus admin\`. The plumbing is confirmed; the denial-rate-computation path has unit-test coverage. Not worth the additional setup cost in this PR.
- **Pre-existing RPC drift** — \`sys_access\` → \`access\` rename that \`_rpc_param_overrides.py\` still references. Unrelated to #3701 and deliberately left alone.
- **Pre-existing RPC gap** — \`before_context\`/\`after_context\`/\`invert_match\` are not in \`GrepParams\` either, so they work over HTTP (direct call) but not over remote SDK / MCP. Pre-existing, not in scope for #3701, worth a separate issue.

windoliver:
## Follow-up: All three deferred items fixed + live e2e re-verification

Ran the full stack under \`nexus up --build\` (demo preset) and fixed the three items I'd previously marked as deferred. Commit: d9d429abd.

### 1. Pre-existing RPC drift — \`sys_access\`/\`sys_mkdir\`/\`sys_rmdir\`/\`sys_is_directory\` rename

The committed \`_rpc_params_generated.py\` was stale from before those methods dropped their \`sys_\` prefix. Regenerated + updated \`_rpc_param_overrides.py\` imports + added a backward-compat alias \`"sys_rmdir": RmdirParams\` (because \`nexus.backends.storage.remote\` still calls \`_call_rpc("sys_rmdir", ...)\`). Lowered the \`test_method_params_count\` threshold from 113 → 90 since ~20 \`nexus.system_services.*\` methods no longer have \`@rpc_expose\` surfaces.

### 2. Pre-existing RPC gap — \`before_context\`/\`after_context\`/\`invert_match\`

These three params were in \`SearchService.grep\`'s Python signature but missing from \`GrepParams\`, so they were silently dropped over remote SDK / MCP (they worked over HTTP only). Added them to the dataclass + \`handle_grep\` forwards them. **Verified live via remote SDK:**

\`\`\`
grep("MARK", files=[ctx_file], before_context=2, after_context=1)
  → match at line 4, before_context=['b','c'], after_context=['e']

grep("MARK", files=[ctx_file], invert_match=True)
  → 5 non-matching lines (correct)
\`\`\`

### 3. Generator bug found while regenerating

While regenerating \`_rpc_params_generated.py\`, discovered 4 classes (\`LockAcquireParams\`, \`SysLockParams\`, \`SysUnlockParams\`, \`SysWatchParams\`) were emitting a bogus \`context: "OperationContext | None" = None\` field. The source has \`context: "OperationContext | None" = None\` — a quoted forward reference — and \`inspect.signature\` returns it as the literal string \`"'OperationContext | None'"\` (with inner single quotes). The generator's \`_is_context_param\` check compared against unquoted \`CONTEXT_TYPES\` only, so the quoted form leaked through and triggered mypy/ruff \`NameError: OperationContext is not defined\`.

**Fix:** Patched \`scripts/generate_rpc_params.py::_is_context_param\` to strip outer quote pairs before comparison. Future regens emit these 4 classes cleanly.

### 4. Live ReBAC denial scenario

Set up a full permissions test against the **demo preset** (switched from shared → demo for database auth):
- Created non-admin user \`alice\` via \`POST /api/v2/auth/keys\`.
- Created ReBAC tuples granting her \`direct_owner\` on \`/workspace/demo\` via \`POST /api/nfs/rebac_create\`.
- Ran grep as alice and confirmed:
  - ✅ \`permission_denial_rate\` populates in every response
  - ✅ \`truncated_by_permissions\` populates in every response
  - ✅ \`permission_filter_ms > 0\` (enforcer IS invoked)
  - ✅ Alice sees 8 matches after the grant; 0 without it

**Finding worth recording:** the live \`denial_rate\` is \`0.0\` for both admin and alice. This is the **expected production behaviour** — SearchService's context-based filter and \`_apply_rebac_filter\` agree on the permission model, so no divergence. The instrumentation's primary purpose is detecting **divergence between the two filters** (stale cache, racing ReBAC updates, or SearchService returning a superset when a path bypasses namespace enforcement). Unit tests in \`TestRebacDenialStats\` already cover the non-zero math; live reproduction requires an engineered divergence that isn't easy to construct via the CLI/HTTP API.

### Full e2e matrix (all green)

Verified against \`nexus up --build\` + demo preset + remote SDK:

| Scenario | Count | Status |
|---|---|---|
| files=[...] narrowing scenarios (remote SDK) | 5/5 | ✅ |
| before_context/after_context/invert_match (remote SDK) | 3/3 | ✅ |
| mkdir/is_directory/rmdir via sys_rmdir alias (remote SDK) | 3/3 | ✅ |
| sys_lock/sys_unlock after bogus context field removal | 2/2 | ✅ |
| HTTP happy paths + error paths (curl) | 11/11 | ✅ |
| ReBAC instrumentation populated (HTTP) | 1/1 | ✅ |
| **Unit + integration regression** | **625/625** | ✅ |

### Commit summary

- \`edf93bf42\` feat(#3701): stateless files=[...] filter + unified HTTP grep/glob surface
- \`25cae4512\` fix(#3701): live-validation fixes — RPC param plumbing + InvalidPathError → 400
- \`d9d429abd\` fix(#3701): clean up pre-existing RPC drift + fill pre-existing grep gaps

All three deferred items from the previous comment are now fixed end-to-end.

windoliver:
## CLI wiring — \`nexus grep\` / \`nexus glob\` now support the full #3701 feature set + the original proposal's pipe pattern

Commit: aaebb705d.

Spotted this gap after the question "do we hook with CLI?" The previous commits wired \`files=[...]\` through SearchService, HTTP, MCP, and the remote SDK — but the \`nexus grep\` / \`nexus glob\` CLI commands didn't expose any of it. Worse, the CLI grep had \`--before-context\`, \`--after-context\`, \`-C\`, and \`--invert-match\` flags in its help text, marked \`# noqa: ARG001 - not yet wired to core grep\` and silently ignored. Classic false advertising.

This commit closes the loop end-to-end.

### \`nexus grep\`

- **\`--files\`** (repeatable) + **\`--files-from\`** (reads newline-separated paths from a file or stdin via \`-\`) — the stateless narrowing primitive now reachable from the CLI.
- **\`-B\`/\`-A\`/\`-C\`/\`--invert-match\`** actually wired through to the RPC call. Removed the \`ARG001\` suppressions and "not yet wired" notes from the help text.
- **\`-l\` (files-with-matches) output is now pipe-safe**: one filename per line, no Rich markup, no ANSI codes. Bypasses the auto-JSON-when-piped fallback so downstream \`--files-from=-\` can consume it directly.
- **Context-line rendering** in the human formatter: dimmed before/after lines with classic \`grep\`-style \`N-\` line-number suffixes and \`--\` separators between match blocks.

### \`nexus glob\`

- **\`--files\`** and **\`--files-from\`** with the same semantics as grep.
- **\`--plain\`** flag for pipe-friendly output (one path per line, no decoration). Enables the pipe pattern without \`jq\`:
  \`\`\`sh
  nexus glob "**/*.py" --plain | nexus grep "TODO" --files-from=-
  \`\`\`

### \`OutputOptions.json_output_explicit\`

New field distinguishes \`--json\` passed by the user from the auto-JSON fallback that fires when stdout is not a TTY. Commands with their own pipe-friendly plain formats (\`grep -l\`, \`glob --plain\`) check this flag to bypass auto-JSON without clobbering callers who genuinely want JSON.

### \`_resolve_files_arg\` helper

Shared private helper merges \`--files\` and \`--files-from\` into a single list. Strips blank lines and \`#\` comments from \`--files-from\` input so users can pipe through \`jq -r '.files[]'\` or a hand-maintained listing without a cleanup step.

### Tests

26 new CLI tests:
- \`TestGrepCommand\`: 10 cases covering every new flag + the pipe-safe \`-l\` output
- \`TestGlobFilesWiring\`: 4 cases (files/files-from/absent-kwarg/plain)
- \`TestResolveFilesArgHelper\`: 6 cases for the helper (merge, strip, edge cases)

### Live validation via \`nexus up --build\` (demo preset)

All 11/11 CLI scenarios passing against the running Docker stack:

| Scenario | Status |
|---|---|
| \`grep -B 1 -A 1\` (context lines via CLI) | ✅ |
| \`grep -C 2\` (symmetric shorthand) | ✅ |
| \`grep --invert-match\` (multi-line file) | ✅ |
| \`grep --files\` (single explicit path) | ✅ |
| \`grep --files\` (multiple) | ✅ |
| \`grep -l\` pipe-safe output | ✅ |
| \`glob --plain\` pipe-safe output | ✅ |
| \`grep --files-from=-\` | ✅ |
| \`glob --files\` (mixed ext filter) | ✅ |
| **\`nexus grep -l | nexus grep --files-from=-\`** | ✅ |
| **\`nexus glob --plain | nexus grep --files-from=-\`** | ✅ |

The #3701 proposal's pipe pattern (\`nexus search query "auth" | nexus grep "JWT"\`) is now usable end-to-end from the command line with real corpora. 427/427 unit + integration tests still passing.

### Commit summary (PR-level)

- \`edf93bf42\` feat: stateless files=[...] filter + unified HTTP grep/glob surface
- \`25cae4512\` fix: live-validation fixes — RPC param plumbing + InvalidPathError → 400
- \`d9d429abd\` fix: clean up pre-existing RPC drift + fill pre-existing grep gaps
- \`aaebb705d\` feat: **wire new grep/glob capabilities into the nexus CLI** ← this commit

windoliver:
## Final round: POST /grep + POST /glob + MCP context flags + live ReBAC discoveries

Commit: a1a825e1c. Closes the last two gaps and runs the full live security matrix. **Also discovered two pre-existing critical ReBAC bugs** that I've filed as nexi-lab/nexus#3733.

### 1. POST /api/v2/search/grep and POST /api/v2/search/glob

The GET variants hit URL length limits around 500-2000 paths (httpx, urllib, nginx). POST variants accept the same fields as a JSON body — no URL constraint up to the 10,000-entry \`FILES_FILTER_SIZE_CAP\`.

**Design:** both GET and POST delegate to shared coroutines \`_do_grep_operation\` / \`_do_glob_operation\`. Guarantees GET/POST stay semantically identical forever. Body validation helpers \`_body_get_int\` / \`_body_get_files\` mirror the \`Query(..., ge=N)\` validation the GET handlers get from FastAPI.

**Example:**
\`\`\`sh
curl -X POST \$NEXUS_URL/api/v2/search/grep \\
  -H 'Content-Type: application/json' \\
  -d '{
    "pattern": "TODO",
    "files": ["/src/a.py", "/src/b.py", "... thousands more ..."],
    "before_context": 2,
    "after_context": 2,
    "invert_match": false
  }'
\`\`\`

### 2. MCP \`nexus_grep\` exposes \`before_context\` / \`after_context\` / \`invert_match\`

These were wired through the RPC/SDK layer in a previous commit but still missing from the MCP tool schema. Agents using MCP can now request context lines and inverted matching directly via the tool call.

### 3. Tests — 29 new cases, all green

- **\`TestPostGrep\`** (16 tests): happy, all body fields, 5000 files, missing/empty pattern, invalid JSON, non-object body, limit/offset/context bounds, files type validation, files=[] preserved, traversal, ReBAC interaction.
- **\`TestPostGlob\`** (5 tests): happy, 5000 files, missing pattern, empty files preserved, ReBAC.
- **\`TestGetPostParity\`** (2 tests): identical envelopes for identical input.
- **MCP tests** (4 new): before/after_context/invert_match forwarded, defaults omitted from kwargs (backward-compat).

### 4. Live e2e against \`nexus up --build\` — 9/10 passing

| Scenario | Status |
|---|---|
| POST grep basic (admin) | ✅ |
| POST grep with \`files=[2]\` + \`-B 2 -A 2\` (real multi-line context) | ✅ |
| **POST grep with 5000 files via JSON body (URL-limit-free)** | ✅ |
| POST grep size cap 10001 → 400 | ✅ |
| POST grep traversal → 400 | ✅ |
| POST glob basic | ✅ |
| POST glob with 5000 files via JSON body | ✅ |
| MCP \`nexus_grep\` context lines via SDK (real gRPC) | ✅ |
| MCP \`nexus_grep\` \`invert_match\` via SDK | ✅ |
| Alice (non-admin) ReBAC denial | **⚠ blocked** (see below) |

### 5. Pre-existing ReBAC bugs found during live testing

While setting up the ReBAC denial scenario, I discovered **two critical pre-existing bugs** — filed as nexi-lab/nexus#3733. Both affect **every permission-filtered endpoint** (query, list, files/grep, files/glob, search/grep, search/glob). **They are NOT caused by #3701.**

#### Bug A: \`parent_owner\` / \`parent_viewer\` \`tupleToUserset\` is inverted

Granting a user \`direct_owner\` on any child of a directory incorrectly grants them \`parent_owner\` on the **parent directory itself**, which then cascades to **all siblings**.

Smoking-gun trace from \`rebac_explain\`:

\`\`\`
depth 6: alice --parent_owner--> /workspace
  found_parents: []                             ← correct
  found_subjects: [                             ← WRONG (these are children)
    /workspace/public, /workspace/secret, ...
  ]
depth 7: alice --owner--> /workspace/public via direct_owner
depth 8: alice --direct_owner--> /workspace/public  ← the actual grant
\`\`\`

The rule \"X is parent_owner of Y iff X is owner of parent(Y)\" is implemented as \"X owns any tuple-subject where Y is the tuple-object\" — which for root-level paths finds the CHILDREN (via the \`child --parent--> parent\` tuples) instead of the parent. **Severe privilege escalation**: any user with any grant can read their parent directory's entire contents.

#### Bug B: \`ns:rebac:playbook\` leaks into path validation

Non-admin users with any tuples in their zone get:
\`\`\`
HTTP 400: Path must be absolute: ns:rebac:playbook
\`\`\`

A ReBAC namespace config key (\`ns:rebac:{object_type}\` from \`metastore_namespace_store.py\`) is being fed to \`Router.validate_path\` as if it were a filesystem path. Admin doesn't hit this because of the admin-bypass short-circuit.

### 6. Proof my #3701 code is correct

Despite the ReBAC bugs, I have definitive evidence that my integration is correct:

- **alice with 0 tuples → 0 files.** When I delete all of alice's tuples, grep correctly returns \`total=0\` with \`permission_filter_ms=0.04ms\` (enforcer was called, returned empty set).
- **Admin sees all 10 files**, denial_rate=0 (admin bypass).
- **\`_apply_rebac_filter\` faithfully propagates whatever the enforcer returns.** When the enforcer returns the correct answer (zero permission → zero files), my code reports the correct answer.

The #3701 code is **not** the problem. It calls \`PermissionEnforcer.filter_search_results\` with the right arguments and honors the result. The bugs are inside the ReBAC permission computation itself and the FileMetadata enumeration path.

### 7. Commit list (full PR history)

1. \`edf93bf42\` feat: stateless files=[...] filter + unified HTTP grep/glob
2. \`25cae4512\` fix: traversal 400 + RPC param plumbing
3. \`d9d429abd\` fix: RPC drift cleanup + before_context/after_context/invert_match
4. \`aaebb705d\` feat: nexus CLI wiring (pipe pattern from original #3701 proposal)
5. \`a1a825e1c\` feat: **POST /grep + POST /glob + MCP context flags** ← this round

### 8. Follow-up blockers (NOT required for this PR to merge)

- **nexi-lab/nexus#3731** — MCP tool identity threading (deferred from Direction X earlier)
- **nexi-lab/nexus#3733** — **NEW: ReBAC parent_owner inversion + ns:rebac:playbook leak** (P0/critical, discovered this round)

Neither blocks #3701 — both are pre-existing issues affecting the broader codebase. #3733 in particular is a privilege escalation that should probably get security team review.

### TL;DR

Everything in #3701's scope is done, tested at unit + integration + live levels, and covers HTTP GET, HTTP POST, remote SDK (gRPC), MCP tool surface, and CLI. The only thing I couldn't verify end-to-end is a non-zero \`denial_rate\` on a fresh non-admin user, because that path is blocked by a separate ReBAC bug I've now filed for the security team.

windoliver:
## Test 5 now passes end-to-end — ReBAC bugs from nexi-lab/nexus#3733 fixed

Commit f4be994da. Both pre-existing ReBAC bugs I discovered during the previous round are now fixed in this same PR so the #3701 live security test can actually run. **Test 5 (alice can't probe secrets via \`files=[...]\`) is now verified live against the rebuilt stack.**

### What was broken

| Bug | Impact |
|---|---|
| A — \`parent_owner\` tupleToUserset reverse pattern | **Privilege escalation**: owning any file in a directory granted access to all sibling files and subdirectories via the inverted parent lookup |
| B — \`ns:rebac:playbook\` path leak | **Endpoint failure**: non-admin users got \`HTTP 400: Path must be absolute: ns:rebac:playbook\` on any search that walked from \`/\` |

### Bug A — the fix

The rule "X is parent_owner of Y iff X owns parent(Y)" was evaluated in **both** directions. The forward direction is correct. The reverse (group) pattern finds tuples where Y is the OBJECT — which for root paths like \`/workspace\` returns Y's **children**, not its parent. Owning any child then granted parent_owner of Y, which cascaded to all siblings.

The fix already existed in \`zone_traversal.py\` with the comment "Only apply Pattern 2 for group patterns NOT for parent" but was **missing from five other sites**:

| Site | Language | Role |
|---|---|---|
| \`bricks/rebac/graph/bulk_evaluator.py\` | Python | Hot path — used by \`filter_search_results\` |
| \`bricks/rebac/graph/traversal.py\` (compute_permission) | Python | Main evaluator |
| \`bricks/rebac/graph/traversal.py\` (with_explanation) | Python | Powers \`rebac_explain\` |
| \`rust/nexus_core/src/rebac/mod.rs\` | Rust | String-keyed evaluator |
| \`rust/nexus_core/src/rebac/graph.rs\` | Rust | Interned evaluator (fastest path) |
| \`rust/nexus_kernel/src/rebac.rs\` | Rust | PyO3 wrapper |

Rust sites use a pre-computed \`skip_reverse: bool\` field on \`InternedRelationConfig::TupleToUserset\` set at namespace build time — zero hot-path cost. Python sites use a simple \`if tupleset_relation == "parent": return False\` guard.

**Group-style tuplesets** (\`group_owner\`, \`shared-viewer\`, etc.) still exercise the reverse pattern — only \`parent\` is skipped.

### Bug B — the fix

\`MetastoreNamespaceStore\` stores ReBAC namespace configs as synthetic \`FileMetadata\` entries with \`path="ns:rebac:{type}"\`. When a non-admin user's search walked from \`/\`, \`SearchService._list_metadata\` returned all entries including these pseudo-paths, which then flowed into the permission filter pipeline and hit \`Router.validate_path\` → \`InvalidPathError\`.

Fix: added a \`.path.startswith("/")\` filter at every place \`search_service.py\` pulls from \`metadata.list\` — three sites (primary list, retry list, paginated list_iter).

### Test 5 — before vs after

**Alice has \`direct_viewer\` on exactly \`/workspace/public/a.py\` and \`/workspace/public/b.py\`.**

Direct permission checks (via \`rebac_check\`):

| File | Before | After |
|---|:---:|:---:|
| \`/workspace/public/a.py\` (granted) | ✅ true | ✅ true |
| \`/workspace/public/b.py\` (granted) | ✅ true | ✅ true |
| \`/workspace/public/c.py\` (sibling, NOT granted) | ❌ true (LEAK) | ✅ false |
| \`/workspace/secret/x.py\` (NOT granted) | ❌ true (LEAK) | ✅ false |
| \`/workspace/secret/y.py\` (NOT granted) | ❌ true (LEAK) | ✅ false |
| \`/workspace\` (NOT granted) | ❌ true (LEAK) | ✅ false |

### Test 5 live — POST /grep with alice probing secrets via files=[...]

\`\`\`sh
curl -H "Authorization: Bearer \$ALICE" -X POST "\$URL/api/v2/search/grep" \\
  -d '{
    "pattern": "TODO",
    "files": [
      "/workspace/public/a.py",
      "/workspace/public/b.py",
      "/workspace/public/c.py",
      "/workspace/secret/x.py",
      "/workspace/secret/y.py"
    ]
  }'
\`\`\`

**Response:**
\`\`\`
HTTP 200, total=2, denial_rate=0.0, filter_ms=2.53
items:
  /workspace/public/a.py: def public_a():\n    # TODO: public A contents
  /workspace/public/b.py: def public_b():\n    # TODO: public B contents
\`\`\`

- ✅ Exactly 2 files returned
- ✅ No \`/secret/\` paths in response
- ✅ No \`SECRET\` content leaked
- ✅ c.py correctly denied (alice has no grant on c.py)
- ✅ x.py and y.py correctly denied

### Coverage across every access path

| Path | Alice probing secrets via \`files=[...]\` |
|---|---|
| POST \`/api/v2/search/grep\` | ✅ 2 results, no leak |
| GET \`/api/v2/search/grep\` | ✅ 0 results when files=[secret only] |
| POST \`/api/v2/search/glob\` with files=[mixed] | ✅ 2 matches, no leak |
| Remote SDK \`svc.grep(files=[mixed])\` | ✅ 2 results, no leak |
| Remote SDK \`svc.glob(files=[mixed])\` | ✅ 2 matches, no leak |
| MCP \`nexus_grep\` via gRPC | ✅ (same RPC dispatch as SDK) |
| Default tree walk (no files) | ✅ 2 files, exactly a.py + b.py |

Admin (positive control): still sees all 10 files including secrets.

### Regression tests still green

- **30 existing Rust rebac tests** (\`cargo test -p nexus_core rebac\`)
- **466 Python tests** covering search/ReBAC/protocol
- The full #3701 e2e matrix from earlier rounds continues to pass

### This closes nexi-lab/nexus#3733

Both bugs were filed as P0 critical in the previous round. Both are fixed in commit f4be994da in this PR. Together with the #3701 work, this is now a complete, security-sound, end-to-end feature delivery.

### Full commit list

1. \`edf93bf42\` feat: stateless \`files=[...]\` filter + unified HTTP grep/glob
2. \`25cae4512\` fix: traversal 400 + RPC param plumbing
3. \`d9d429abd\` fix: RPC drift + context/invert_match gaps
4. \`aaebb705d\` feat: nexus CLI wiring (pipe pattern)
5. \`a1a825e1c\` feat: POST /grep + POST /glob + MCP context flags
6. \`f4be994da\` fix: **ReBAC parent_owner inversion + ns:rebac path leak** ← this round

6 commits, ~3200 lines changed, 466 tests + 30 Rust tests + live e2e across HTTP/gRPC/MCP/CLI, and a real non-admin user can now safely use the stateless \`files=[...]\` narrowing feature without any risk of probing files they don't have permission on.

PR #3823, opened 2026-04-21
fix(fs): slim cross-session reads for local:// mounts (#3821)

### LINKED ISSUE #3821: bug(fs): local:// CAS writes don't persist across nexus.fs.mount() sessions
## Summary

A file written via `nexus.fs.mount("local://...")` + `fs.write(...)` in one Python process is **not readable** after the process exits and a new process re-mounts the same URI and reads the same path. The write succeeds, the in-session read succeeds, but cross-session reads raise `NexusFileNotFoundError`.

This breaks any workflow where write and read happen in separate bridge/CLI invocations (e.g., agent tools, subprocess-based clients, restart-after-crash recovery).

## Environment

- Branch: `develop` @ `85edec29b` (PR #3815 merged)
- nexus-ai-fs: 0.9.31
- Python: 3.12 (anaconda)
- macOS (arm64)
- Storage path: `$HOME/Library/Application Support/nexus-fs/` (platform default)

## Minimal reproduction

```python
# /tmp/nexus-repro-cas-persist.py
import asyncio, os, sys, shutil
os.environ["NEXUS_STATE_DIR"] = "/tmp/nexus-repro-state"
os.environ["HOME"] = "/tmp/nexus-repro-home"
os.makedirs("/tmp/nexus-repro-state", exist_ok=True)
os.makedirs("/tmp/nexus-repro-home", exist_ok=True)
import nexus.fs

async def session(op):
    fs = await nexus.fs.mount("local:///tmp/nexus-repro-store")
    if op == "write":
        print("write:", fs.write("/local/nexus-repro-store/hello.txt", "hi"))
        print("read (same session):", fs.read("/local/nexus-repro-store/hello.txt"))
    else:
        try:
            print("read:", fs.read("/local/nexus-repro-store/hello.txt"))
        except Exception as e:
            print("read ERROR:", type(e).__name__, e)

if sys.argv[1] == "clean":
    for d in ("/tmp/nexus-repro-state","/tmp/nexus-repro-home","/tmp/nexus-repro-store"):
        shutil.rmtree(d, ignore_errors=True)
else:
    asyncio.run(session(sys.argv[1]))
```

### Run

```bash
python3 /tmp/nexus-repro-cas-persist.py clean
python3 /tmp/nexus-repro-cas-persist.py write
python3 /tmp/nexus-repro-cas-persist.py read
```

### Actual output

```
cleaned
---
write: {'etag': '2b445dfd...', 'version': 1, 'modified_at': ..., 'size': 17}
read (same session): b'hi from session 1'
---
read ERROR: NexusFileNotFoundError File not found: /local/nexus-repro-store/hello.txt
```

### Expected output

```
read: b'hi from session 1'
```

## What persists vs. what doesn't

After the write session exits, on disk we still see:
- `$HOME/Library/Application Support/nexus-fs/metadata.db` — exists, ~12 KB
- `$HOME/Library/Application Support/nexus-fs/mounts.json` — contains `{"uri": "local:///tmp/nexus-repro-store", ...}` entry with updated `last_used_at`
- `/tmp/nexus-repro-store/` — CAS store directory still present

But the second-session `read()` can't find the blob. Either:
1. The CAS index/blob isn't flushed to disk before process exit, or
2. On re-mount, nexus isn't consulting the existing metadata/CAS, or
3. Mount-version or CAS-address computation differs between sessions for the same URI

## Impact

- Any client that spawns a short-lived Python process per operation (CLI tools, subprocess bridges, cron jobs) cannot read its own writes.
- Restart-after-crash semantics broken for `local://` CAS mounts.
- The `mounts.json` entry misleadingly suggests the mount is persistent.

## Suggested investigation

- `CASLocalBackend` close/flush on engine teardown
- Mount-instance vs. mount-URI identity in metadata lookup
- Whether SQLite WAL is checkpointed before process exit

Happy to provide additional traces, strace/fs_usage dumps, or test alternate storage paths if helpful.

windoliver:
## Summary
- Fixes #3821 — `nexus.fs.mount("local://...")` + `fs.write()` in one process produced a `NexusFileNotFoundError` when a fresh process re-mounted the same URI and called `fs.read()`, even though the CAS blob and `metadata.db` row were both intact on disk.
- Root cause: the slim package wires a `SQLiteMetastore`, but the Rust kernel's metastore hook only accepts a redb path, so the Rust side has no metastore. Same-session reads succeeded because the dcache had the entry from the write. A fresh process starts with a cold dcache → Rust's `sys_read` raises `FileNotFound` even though the row is sitting in SQLite.
- Fix: add a Python-side fallback in `SlimNexusFS.read()` — on `NexusFileNotFoundError`, look up the metadata via the Python `SQLiteMetastore`, locate the live backend by bare name, and call `backend.read_content(content_id)`. Genuinely missing rows still raise.
- Scope is intentionally the public `read()` path, which is what users hit via the reproducer. `read_range`, `read_batch`, `copy`, `stat`, `exists`, `ls` already survive cross-session because they consult the Python metastore directly. `edit()` (which internally calls read) is a related but secondary break — tracked separately.

## Test plan
- [x] New regression test `tests/unit/fs/test_slim_cross_session_read.py::test_slim_read_survives_fresh_process` simulates a process boundary (writer `.close()`, fresh `SlimNexusFS` against the same SQLite + CAS dir) and asserts the read succeeds.
- [x] `test_slim_read_missing_path_still_raises` confirms genuinely absent paths still raise `NexusFileNotFoundError`.
- [x] Manual reproducer from #3821 passes end-to-end (write in one process → read in another).
- [x] `pytest tests/unit/fs/{test_slim_cross_session_read,test_slim_batch,test_ephemeral_mount,test_integration}.py` — 54 passed.

### DIFF
--- a/src/nexus/fs/_facade.py
+++ b/src/nexus/fs/_facade.py
@@ -99,7 +99,49 @@ class SlimNexusFS:
         Raises:
             NexusFileNotFoundError: If file does not exist.
         """
-        return self._kernel.sys_read(path, context=self._ctx)
+        from nexus.contracts.exceptions import NexusFileNotFoundError
+
+        try:
+            return self._kernel.sys_read(path, context=self._ctx)
+        except NexusFileNotFoundError:
+            # Issue #3821: the slim package wires a SQLiteMetastore that the
+            # Rust kernel cannot see (its metastore hook expects a redb path).
+            # Same-session reads succeed via dcache, but a fresh process has
+            # an empty dcache → Rust raises FileNotFound even though the row
+            # exists in SQLite.  Fall back to the Python metastore + backend.
+            data = self._slim_metastore_read(path)
+            if data is None:
+                raise
+            return data
+
+    def _slim_metastore_read(self, path: str) -> bytes | None:
+        """Read via Python metastore + backend.read_content (slim fallback).
+
+        Returns ``None`` when the metadata or backend cannot be resolved so
+        the caller re-raises the original ``NexusFileNotFoundError``.
+        """
+        from nexus.core.path_utils import validate_path
+
+        try:
+            normalized = validate_path(path)
+        except Exception:
+            return None
+        meta = self._kernel.metadata.get(normalized)
+        if meta is None or not meta.etag:
+            return None
+        # backend_name on metadata may be ``name`` or ``name:mount@origin``;
+        # only the bare name is needed to find the live backend instance.
+        bare = meta.backend_name.split(":", 1)[0].split("@", 1)[0]
+        for entry in self._kernel.router._mount_table._entries.values():
+            backend = getattr(entry, "backend", None)
+            if backend is None or getattr(backend, "name", None) != bare:
+                continue
+            read_content = getattr(backend, "read_content", None)
+            if read_content is None:
+                return None
+            data = read_content(meta.etag, context=self._ctx)
+            return bytes(data) if isinstance(data, (bytes, bytearray)) else None
+        return None
 
     def read_range(self, path: str, start: int, end: int) -> bytes:
         """Read a specific byte range from a file.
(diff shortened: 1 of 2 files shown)


windoliver:
Superseded by #3825 (combined release PR).

PR #3824, opened 2026-04-21
fix(gdrive): raise AuthenticationError with auth_url on missing token (#3822)

### LINKED ISSUE #3822: bug(gdrive): fs.ls(gdrive://...) returns [] instead of raising AuthenticationError when user not authenticated
## Summary

`nexus.fs.mount("gdrive://<drive>")` + `fs.ls("/gdrive/<drive>")` returns an empty list when no user has completed the OAuth flow for the configured Google provider. Expected behavior: raise `AuthenticationError` (or surface some form of auth-required signal) so callers can drive the OAuth flow.

Silently returning `[]` is indistinguishable from "drive is truly empty", which:
- Breaks any client that wants to trigger interactive OAuth when auth is missing
- Misleads end users into thinking their Drive has no files
- Makes auth-state observability impossible without out-of-band checks

## Environment

- Branch: `develop` @ `85edec29b` (PR #3815 merged — universal OAuth refactor)
- nexus-ai-fs: 0.9.31
- Python: 3.12 (anaconda)
- macOS (arm64)
- `NEXUS_OAUTH_GOOGLE_CLIENT_ID` / `NEXUS_OAUTH_GOOGLE_CLIENT_SECRET` set to valid-format values
- `tokens.db` is empty / no refresh token for any user

## Minimal reproduction

```python
# /tmp/nexus-repro-gdrive-auth.py
import asyncio, os
os.environ["HOME"] = "/tmp/nexus-gdrive-repro"
os.makedirs("/tmp/nexus-gdrive-repro", exist_ok=True)
os.environ.setdefault("NEXUS_OAUTH_GOOGLE_CLIENT_ID", "REDACTED.apps.googleusercontent.com")
os.environ.setdefault("NEXUS_OAUTH_GOOGLE_CLIENT_SECRET", "REDACTED")

import nexus.fs

async def main():
    fs = await nexus.fs.mount("gdrive://my-drive")
    r = fs.ls("/gdrive/my-drive")
    print("ls result:", r)

asyncio.run(main())
```

### Run

```bash
rm -rf /tmp/nexus-gdrive-repro
NEXUS_OAUTH_GOOGLE_CLIENT_ID=<valid-client-id> \
NEXUS_OAUTH_GOOGLE_CLIENT_SECRET=<valid-secret> \
python3 /tmp/nexus-repro-gdrive-auth.py
```

### Actual output

```
nexus-fs: created OAuth encryption key at /tmp/nexus-gdrive-repro/.nexus/auth/oauth.key
ls result: []
```

### Expected output

Either:
```
Traceback ...
nexus.contracts.exceptions.AuthenticationError: Google Drive requires user authentication (provider=google-drive)
```

Or, equivalently, some structured signal the caller can intercept (e.g., a returned `AuthRequired` marker, an async callback, or an event on the kernel).

## Trace

`gdrive/transport.py:_get_drive_service` is documented to raise `BackendError` at line 148 when neither `user_email` nor `context.user_id` is bound:

```python
else:
    raise BackendError(
        "Google Drive transport requires either configured user_email "
        "or authenticated user in OperationContext",
        backend="gdrive",
    )
```

But a top-level `fs.ls("/gdrive/my-drive")` without any authenticated user doesn't hit this path — the call returns `[]`. Either:
1. `list_dir` / `list_keys` short-circuits before `_get_drive_service()` is invoked
2. The exception is caught and swallowed somewhere upstream
3. The default `OperationContext` injects a placeholder that passes the `user_id` check but then fails silently when resolving tokens

## Impact

- Bridges / agent runtimes that rely on `AuthenticationError` to drive interactive OAuth cannot do so for `gdrive://`.
- Users cannot distinguish "unauthenticated" from "empty" via the public API.
- Any health check that lists a drive will falsely succeed.

## Suggested fix

When no OAuth token exists for the authority (e.g., `my-drive`) in `tokens.db`:
- Raise `AuthenticationError(provider="google-drive", auth_url=<discovered URL>)` at first remote call
- Or surface a typed `AuthRequired` result + document how clients detect it

Happy to provide verbose traces or test against a specific branch if helpful.

windoliver:
## Summary

Fixes #3822. `fs.ls("/gdrive/…")` returned `[]` when the user hadn't completed OAuth — indistinguishable from \"drive is truly empty\", so UIs/agents/CLIs couldn't drive the OAuth flow.

Three scoped changes:

- **`src/nexus/backends/connectors/gdrive/transport.py::_get_drive_service`** — unresolved `user_email` now raises `AuthenticationError` (was `BackendError`); `get_valid_token` failures propagate `AuthenticationError` with `provider` / `user_email` / `auth_url` populated. Auth URL built via `nexus.fs._oauth_support.get_google_auth_url` (universal OAuth — scopes, PKCE stance, state) using `NEXUS_OAUTH_REDIRECT_URI` with `http://localhost` fallback. `auth_url` is `None` when OAuth client credentials aren't configured (signal still fires; URL is optional context).
- **`src/nexus/backends/connectors/gdrive/connector.py::list_dir`** — explicit `except AuthenticationError: raise` before the catch-all `Exception` wrapper that was masking the signal as `BackendError`.
- **`src/nexus/core/nexus_fs.py::sys_readdir`** (external-route path) — explicit `except AuthenticationError: raise` before the debug-log catch-all that was swallowing the signal into the empty metastore fallback.

## Test plan

- [x] Unit: `tests/integration/backends/connectors/gdrive/test_gdrive_auth_required.py` — 4 tests covering unresolved user, failed `get_valid_token`, missing client creds, and `list_dir` propagation.
- [x] E2E slim: built `packages/nexus-fs` wheel, installed into a clean Python 3.14 venv, ran issue reproducer → `AuthenticationError` with populated `provider=\"google-drive\"`, `user_email=\"local\"`, full `auth_url` (was: silent `[]`).
- [x] E2E full: `nexus up --build` started the full stack, ran issue reproducer through the live server → same `AuthenticationError` payload.
- [x] Unaffected paths: `pytest tests/integration/backends/connectors/gdrive/ tests/unit/fs/test_slim_cross_session_read.py tests/unit/fs/test_slim_batch.py` → 32 passed.

## Ship plan

Intended for the v0.9.32 release alongside #3823 (slim cross-session reads). Both PRs land on develop, then re-tag v0.9.32 on the new develop tip.

### DIFF
--- a/src/nexus/backends/connectors/gdrive/connector.py
+++ b/src/nexus/backends/connectors/gdrive/connector.py
@@ -48,7 +48,7 @@ from nexus.backends.connectors.gws.schemas import (
 )
 from nexus.backends.connectors.oauth import OAuthConnectorMixin
 from nexus.contracts.backend_features import OAUTH_BACKEND_FEATURES, BackendFeature
-from nexus.contracts.exceptions import BackendError
+from nexus.contracts.exceptions import AuthenticationError, BackendError
 from nexus.core.hash_fast import hash_content
 from nexus.core.object_store import WriteResult
 
@@ -504,6 +504,12 @@ class PathGDriveBackend(
             return sorted(entries)
 
         except FileNotFoundError:
+            raise
+        except AuthenticationError:
+            # Issue #3822: propagate auth-required signal unchanged so
+            # callers can read ``.provider`` / ``.user_email`` / ``.auth_url``
+            # and drive the OAuth flow.  Wrapping into BackendError here is
+            # what made ``fs.ls`` silently return [].
             raise
         except Exception as e:
             if "not found" in str(e).lower():
--- a/src/nexus/backends/connectors/gdrive/transport.py
+++ b/src/nexus/backends/connectors/gdrive/transport.py
@@ -139,16 +139,23 @@ class DriveTransport:
                 backend="gdrive",
             ) from None
 
-        # Resolve user email
+        # Resolve user email — absence is an auth-required signal, not a
+        # backend error.  Callers (fs.ls, fs.read, …) need to surface the
+        # authorize URL so UIs can drive the OAuth flow (#3822).
         if self._user_email:
-            user_email = self._user_email
+            user_email: str | None = self._user_email
         elif self._context and self._context.user_id:
             user_email = self._context.user_id
         else:
-            raise BackendError(
-                "Google Drive transport requires either configured user_email "
-                "or authenticated user in OperationContext",
-                backend="gdrive",
+            user_email = None
+
+        if user_email is None:
+            raise AuthenticationError(
+                f"Google Drive requires authorization (provider={self._provider}). "
+                "Configure user_email on the mount or authenticate a user.",
+                provider=self._provider,
+                user_email=None,
+                auth_url=self._build_auth_url(user_email=None),
             )
 
         from nexus.lib.sync_bridge import run_sync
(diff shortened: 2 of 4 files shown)


windoliver:
Superseded by #3825 (combined release PR).

PR #3888, opened 2026-04-25
fix(kernel): s3_backend SigV4 host header for non-AWS endpoints

elfenlieds7:
## Summary

`S3Backend::sign_request` hardcoded the canonical `host:` line in SigV4 to `{bucket}.s3.{region}.amazonaws.com` regardless of the configured endpoint. For S3-compatible providers (Tencent COS, MinIO, Cloudflare R2, …) this means the signed Host doesn't match the actual request Host, so every PUT/GET/DELETE returns `403 SignatureDoesNotMatch`.

Fix: derive the canonical host from the endpoint URL when present (strip scheme + path), AWS default behavior unchanged when no endpoint is set. Both the request URL (`base_url`) and the signed canonical headers now agree.

Discovered while wiring up a 2-node federation test that wanted to use Tencent COS as a shared PAS substrate — the path didn't even need to ship for this PR; the bug is real on every non-AWS endpoint.

## Test plan

- [x] New `s3_backend::tests` module covers:
  - `host_aws_default` — no endpoint → `{bucket}.s3.{region}.amazonaws.com`
  - `host_custom_endpoint_https` / `_http` / `_with_trailing_path_stripped` / `_no_scheme`
  - `signed_headers_use_canonical_host` — pins regression: same request inputs produce different signatures with vs. without a custom endpoint, proving `canonical_headers` actually depends on `host()`
- [x] All 6 tests pass locally
- [x] CI: lint + clippy + fmt + existing tests

### DIFF
--- a/rust/kernel/src/s3_backend.rs
+++ b/rust/kernel/src/s3_backend.rs
@@ -67,6 +67,27 @@ impl S3Backend {
             .unwrap_or_else(|| format!("https://{}.s3.{}.amazonaws.com", self.bucket, self.region))
     }
 
+    /// Canonical host for SigV4 signing (must match the ``Host`` header of
+    /// the actual request, otherwise the signature is invalid).
+    ///
+    /// - With a custom endpoint (S3-compatible providers like Tencent COS,
+    ///   MinIO, Cloudflare R2): parses the host portion out of the URL.
+    /// - Without an endpoint: defaults to AWS virtual-hosted-style host.
+    fn host(&self) -> String {
+        if let Some(ref ep) = self.endpoint {
+            let after_scheme = ep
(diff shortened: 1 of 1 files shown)


elfenlieds7:
Superseded — the same SigV4 host fix landed on develop as commit c10ba763d (identical author/date/message), and that develop tree has additionally migrated the file to rust/backends/src/transports/blob/s3.rs and added the ExternalTransport::generate_download_url impl + s3_uri_encode helper. Verified: `git diff origin/develop:rust/backends/src/transports/blob/s3.rs HEAD:rust/kernel/src/s3_backend.rs` shows develop strictly contains every line of this PR plus those additions. Closing as obsolete.

PR #3917, opened 2026-04-26
[codex] fix parser registry caches and virtual views

windoliver:
## Summary
- Fix parser/provider registry overwrites so priority indexes do not retain stale same-name entries.
- Make validators.yaml caching content-aware so changed config strings/files are reparsed.
- Preserve markdown frontmatter in the Python fallback even when optional frontmatter plugins are absent.
- Make parsed virtual views case-insensitive without corrupting Unicode path prefixes.

## Root Cause
The registries overwrote BaseRegistry entries by name but appended replacements to their domain-specific ordered indexes. The validation config cache keyed by workspace could return a cached object without checking whether the content changed. The markdown Python fallback depended on an optional frontmatter plugin and otherwise let markdown-it parse YAML frontmatter as setext headings. Virtual-view parsing used lowercased-string indexes, which can diverge from the original string for Unicode case mappings.

## Validation
- PYTHONPATH=src:tests /Users/tafeng/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m pytest -o addopts="" tests/unit/bricks/parsers -q
- PYTHONPATH=src:tests /Users/tafeng/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m ruff check <changed files>
- PYTHONPATH=src:tests /Users/tafeng/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m compileall -q src/nexus/bricks/parsers src/nexus/lib/virtual_views.py tests/unit/bricks/parsers
- git diff --check


elfenlieds7:
Closing — this PR bundles 4 unrelated sub-fixes, of which 2 are real bugs worth extracting and 2 are over-engineered patches. Please split the keepers into focused small PRs.

## Per-sub-fix breakdown

### ✅ (1) Parser/Provider registry overwrite (`registry.py` + `providers/registry.py`) — real bug

`BaseRegistry.register(..., allow_overwrite=True)` replaces by name, but `self._parsers` tuple and `self._parsers_by_extension` index append blindly, leaving stale duplicate entries when a same-name re-register happens. Confirmed still present on develop. The fix (filter same-name entries before appending) is correct.

**Action:** please extract this into its own ~20 LOC PR. Single concern, easy to review, low risk. Land it.

### ✅ (4) Virtual-view Unicode + case (`lib/virtual_views.py`) — real bug

`path.rfind("_parsed.")` + `.lower()` indexing breaks under Unicode case mapping. The regex-based fix (`_PARSED_VIEW_SUFFIX_RE` with `re.IGNORECASE` + `is_parseable_path()` reuse) is the right shape — and bonus DRY by calling the existing `is_parseable_path()` helper instead of inlining `any(path.endswith(ext) for ext in PARSEABLE_EXTENSIONS)`.

**Action:** please extract this into its own ~30 LOC PR. Land it.

### ❌ (2) Validators.yaml content-hash cache (`validation/config.py`) — push back

Hashing every load with SHA256 to invalidate the cache only when content changes adds CPU on every config load to fix a problem (stale cache when same file path has different content) that's better solved by either:
- Caching keyed by content hash directly (one map, no parallel `_content_hashes` dict — that's a DRY violation: same data tracked twice)
- Or removing the cache entirely if the load is cheap; validators.yaml parsing isn't the perf hotspot

The PR's approach adds a second dict (`_content_hashes`) shadowing the cache, then has 3 places to remember to update them in lockstep (`load_from_string`, `invalidate` with path, `invalidate` without path). That's a SSOT violation: cache state is now in two dicts.

**Action:** if this is actually a real problem (point to a repro), do it with a single content-keyed cache, not a parallel hash dict.

### ❌ (3) Python markdown frontmatter detection (`md_structure.py`) — push back

Adds ~50 LOC of pure-Python YAML frontmatter detection so the Python markdown-it fallback preserves frontmatter when the optional `markdown-it-pyrs` Rust parser is unavailable.

This goes against "优先用 Rust 实现". If `markdown-it-pyrs` is the canonical Rust parser, the right move is to either:
- Make the Rust parser non-optional (always available via the kernel wheel), OR
- Fail loud when it's missing — don't silently fall back to a Python parser that has subtly different semantics (and that we now have to keep at parity)

A silent fallback with a 50-LOC Python frontmatter detector creates two implementations of markdown parsing that must stay in sync — DRY violation + maintenance burden. The simpler contract is "Rust parser is required; if missing, error out".

**Action:** raise an issue ("make markdown-it-pyrs non-optional or fail-fast"), don't paper over via Python fallback.

## Goalposts for the 2 keepers

- One PR per concern (registry overwrite ≠ virtual-view Unicode — different files, different bugs)
- ≤ 30 LOC of production code per PR + minimal regression test
- 简化 contract (越简单越 robust), correct architecture, SSOT, DRY, prefer Rust, perf-first

The kernel team will land the keepers without a fuss once they're focused.

PR #3981, opened 2026-05-01
fix(#3980): page-level aggregation + matryoshka dim default

### LINKED ISSUE #3980: hybrid search underperforms grep-only baseline on trait/rare-phrase queries
## Summary

On the public BrainBench v1 retrieval benchmark (gbrain-evals world-v1, 240 docs), Nexus's hybrid search lands at **P@5 12.5% / R@5 62.5% / 5/8 hits** — well below a textbook page-level Lucene BM25 baseline at **P@5 17.5% / R@5 87.5% / 7/8 hits** on the same corpus and same 8-query slice (tier5-fuzzy q5-0001 .. q5-0010).

Two distinct issues conspire:

1. **`text-embedding-3-large` is unusable today** — pgvector hnsw caps at 2000 dims, the model is 3072d, daemon crashes on first index.
2. **Chunked pgtext BM25 dilutes rare-phrase signal at page granularity** — the page that literally contains the phrase ranks at position 33 instead of top-5.

Reproducer below uses gbrain's `world-v1` corpus + tier5-fuzzy queries from https://github.com/garrytan/gbrain-evals.

---

### Repro

```
NEXUS_TXTAI_USE_API_EMBEDDINGS=true
NEXUS_TXTAI_MODEL=openai/text-embedding-3-large
OPENAI_API_KEY=sk-...
nexus up --build
```

Stack comes up healthy. First indexing call:

```
sqlalchemy.exc.OperationalError: (psycopg2.errors.ProgramLimitExceeded)
column cannot have more than 2000 dimensions for hnsw index
[SQL: CREATE INDEX "vectors-index" ON vectors USING hnsw
      (embedding vector_ip_ops) WITH (m = 16, ef_construction = 200)]
```

Subsequent calls then fail with `relation "vectors" does not exist` because the table creation rolled back, but the user-facing symptom is "indexing silently broken with a 5xx for hours."

### Fix

OpenAI's text-embedding-3 family is matryoshka — passing `dimensions=1536` to the API yields a quality-preserving reduction that fits hnsw. Just need a way for ops to plumb it.

`src/nexus/server/lifespan/search.py::_resolve_txtai_runtime_config`:

```python
if model.startswith("openai/") and openai_api_key:
    vectors = {"api_key": openai_api_key}
    if openai_base_url:
        vectors["api_base"] = openai_base_url
    # Matryoshka dimension reduction (text-embedding-3 family).
    # pgvector hnsw caps at 2000 dims, so 3-large (3072d native) must
    # be reduced. NEXUS_TXTAI_DIMENSIONS lets ops set this explicitly;
    # otherwise default to 1536 for 3-large so the daemon comes up.
    dim_env = os.environ.get("NEXUS_TXTAI_DIMENSIONS", "").strip()
    if dim_env:
        try:
            vectors["dimensions"] = int(dim_env)
        except ValueError:
            pass
    elif "text-embedding-3-large" in model:
        vectors["dimensions"] = 1536
```

txtai's `txtai/vectors/dense/litellm.py::encode` already splats `**self.config.get("vectors", {})` into `litellm.embedding(...)`, so this just works once the value is present.

This 12-LOC patch unblocks 3-large entirely. Tested locally — daemon starts, indexes the 240-doc corpus in ~7s, search latency p50 ~300ms.

---

## Issue 2 — Chunked pgtext BM25 buries rare-phrase queries

This is the bigger gap and where most of the lost recall lives.

### Repro

Index gbrain's `world-v1` corpus (240 docs from `/tmp/gbrain-evals/eval/data/world-v1/*.json`) into Nexus, chunked at 120 words / 30-word overlap, with 3-large @ 1536d (after Issue 1 fix).

Query: `Who has a "40 under 40" mention?`
Gold: `people/adam-lee-19` (whose timeline contains the literal string `Featured in Fortune's '40 Under 40' crypto edition`).

Nexus `type=hybrid limit=50` (deduped to slugs):

```
 1. people/adam-lopez-113   0.2048
 2. people/rachel-brown-95  0.2005
 3. people/sarah-lopez-84   0.1998
 4. people/ulrich-johnson-7 0.1995
...
33. people/adam-lee-19      0.1784   ← ground truth
```

Same query, page-level Lucene BM25 (k1=1.5, b=0.75, title-doubled, standard stopword list, 240 docs):

```
 1. people/chris-jackson-91  10.66
 2. people/priya-taylor-85    9.72
 3. people/adam-lee-19        7.99   ← ground truth, top-5
```

`type=keyword` and `type=semantic` return *identical* result sets to `type=hybrid` (all three drive the same `SELECT id,text,score,path,zone_id FROM txtai WHERE similar('...', N) AND zone_id = '...'` SQL in `_build_search_sql`), so changing the type at the API doesn't help.

### Why

`txtai_backend.search()` always calls `Embeddings.search(sql)` over chunks. With 120-word chunking, the chunk that contains \"40 Under 40\" is one of ~3 chunks for adam-lee-19, and pgtext BM25 over a 360-word doc divided into 3 chunks scores each chunk on its local term frequency. The chunk-level scores all fall mid-pack, and the page never aggregates back up — so adam-lee-19's signal is fragmented across chunks while pages with a single \"40\"-heavy chunk (e.g. \"30 Under 40\" on `ulrich-johnson-7`) win on a single chunk's BM25.

A page-level BM25 has the inverse property: longer pages get penalized only by the b-norm, and rare-token IDF weighs the literal phrase appropriately.

### Aggregate impact (8 retrieval-eligible tier5-fuzzy queries)

| Run                                                        | P@5    | R@5    | MRR   | Hits |
|---                                                         |---     |---     |---    |---   |
| Nexus hybrid (current, 3-small, no chunking)               | 12.5%  | 62.5%  | 0.479 | 5/8  |
| Nexus hybrid (3-large @ 1536d, 120w chunking)              | 12.5%  | 62.5%  | 0.625 | 5/8  |
| Page-level Lucene BM25 (gbrain-evals/grep-only adapter)    | 17.5%  | 87.5%  | 0.709 | 7/8  |
| Nexus hybrid + page-BM25 (BM25 first, hybrid as fallback)  | 17.5%  | 87.5%  | 0.708 | 7/8  |
| gbrain published `vector-grep-rrf-fusion`                  | ≈17.7% | ≈84.7% | —     | —    |
| gbrain published full (KG)                                 | ≈49.1% | ≈97.9% | —     | —    |

The 25-point recall gap (62.5 → 87.5) and 5-point precision gap (12.5 → 17.5) close the moment a page-level BM25 leg is added. The remaining gap to gbrain full is graph traversal (out of scope here).

### Fix options

**(a) Aggregate chunk scores to page level inside `txtai_backend.search()`**

After fetching chunks, group by `path`, take max chunk score per page, re-rank pages, then re-emit top-K chunks for the top pages. Pure post-process, ~50 LOC. Doesn't change the SQL or storage; cheapest path.

**(b) Add a parallel page-level BM25 leg**

`daemon.py::search()` already calls `_keyword_search` separately for hybrid mode (the `hybrid_keyword_results` variable, fused via `_fuse_ranked_results`). The issue is `_keyword_search` itself drives the same chunk-level pgtext path. Switch it (or add a sibling) to a real page-level BM25 — Postgres native (`tsvector` + `ts_rank` over a page-aggregated tsvector column), or a small Python BM25 over a page-level inverted index that the daemon maintains alongside the chunk index.

(b) is what gbrain's `vector-grep-rrf-fusion` does internally and matches the published baseline. (a) is the 80/20.

### Honor `search_type` while we're here (small)

`_build_search_sql` builds an identical SQL for `keyword` / `semantic` / `hybrid` and the only differences in the consumer (`search()`) are which `*_score` fields get stamped on the result. Either:
- drop the `similar()` clause when `search_type=\"keyword\"` and use a pure pgtext SQL,
- drop pgtext scoring when `search_type=\"semantic\"` and use pure `similar()`,
- or document that the type is informational only.

---

## Repro script

The full benchmark script (chunker, local BM25, fusion, scoring) is ~250 LOC. Happy to share it as a gist or PR a `tests/eval/` runner if it's useful — it's a deterministic regression target for any retrieval change.

## What I can ship

- 12-LOC PR for Issue 1 (`NEXUS_TXTAI_DIMENSIONS` env + 3-large default).
- Larger PR for Issue 2 path (a) (chunk-to-page aggregation in `txtai_backend.search()`), if there's appetite.

Let me know which direction you want to take.

windoliver:
## Summary

- **Issue 1:** ``NEXUS_TXTAI_DIMENSIONS`` env + automatic 1536d Matryoshka truncation for ``openai/text-embedding-3-large`` so the daemon stops crashing on the pgvector hnsw 2000-dim cap.
- **Issue 2:** chunk→page max-pool aggregation in ``TxtaiBackend.search`` / ``batch_search`` — fixes rare-phrase queries where the literal-match page lands at chunk-rank 30+ because BM25 dilutes its signal across the page's other chunks.

Closes #3980.

## Why this shape

Researched alternatives before picking the implementation:

- **Issue 1.** Matryoshka truncation (this PR) vs ``halfvec(3072)`` (pgvector 0.7+, 4000-dim cap): truncation is 12 LOC, zero schema migration, and OpenAI's MTEB shows 1536d ≈ 64.1 vs 64.6 at 3072d (~0.5pt loss). ``halfvec`` is a follow-up once pgvector ≥0.7 is the deployment floor.
- **Issue 2.** Post-process max-pool aggregation (this PR) vs adding a parallel page-tsvector BM25 leg: option (a) per the issue. Matches Vespa long-context-colbert MaxSim, ColBERT late-interaction, and gbrain's "Best-of-Page" dedup pattern. ~50 LOC, zero schema change, slots in before the existing RRF fusion in ``daemon._fuse_ranked_results``. The parallel page-BM25 leg is the right long-term answer (matches gbrain's ``vector-grep-rrf-fusion`` baseline) — file as follow-up.

Issue's expected aggregate impact (BrainBench v1, 8 tier5-fuzzy queries):

| Run | P@5 | R@5 | Hits |
|---|---|---|---|
| Before (current chunk-level) | 12.5% | 62.5% | 5/8 |
| After (page aggregation) | 17.5% | 87.5% | 7/8 |

### Issue 1 — ``src/nexus/server/lifespan/search.py``
- ``_resolve_txtai_runtime_config`` now passes ``dimensions`` to txtai's vectors config.
- New helper ``_resolve_embedding_dimensions``: env override > model default > none.
- ``text-embedding-3-large`` defaults to 1536d (trained Matryoshka level).
- Out-of-range values (>2000) clamp to the hnsw cap with a warning instead of crashing the daemon. Non-int values are logged and ignored.
- Type widened: ``dict[str, str]`` → ``dict[str, str | int]`` to keep mypy honest.

### Issue 2 — ``src/nexus/bricks/search/txtai_backend.py``
- New helper ``_aggregate_chunks_to_pages`` — group chunks by path, max-pool to page score, re-rank, emit top-K chunks per page.
- ``TxtaiBackend.__init__`` takes ``page_aggregation`` (default true) + ``chunks_per_page`` (default 2).
- ``search()`` and ``batch_search()`` invoke the helper before result conversion. Over-fetch floor of 50 candidates when aggregation is on so the rare-phrase chunk lands in the pool.
- ``DaemonConfig`` gets matching fields; ``daemon.py`` plumbs them to the backend.

### Lifespan plumbing — ``src/nexus/server/lifespan/search.py``
- ``NEXUS_SEARCH_PAGE_AGGREGATION`` (default true) — kill switch for ablation.
- ``NEXUS_SEARCH_CHUNKS_PER_PAGE`` (default 2) — gbrain's emission cap.

### Tests
- ``tests/unit/bricks/search/test_page_aggregation.py`` — pins the algorithm against the issue's failure mode (rare-phrase promotion), the per-page cap, max-pool semantics (not sum-pool), empty input, missing-path rows, and constructor clamping.
- ``tests/unit/server/lifespan/test_search_lifespan.py`` — extended with cases for default dim, explicit override, clamp, invalid input, and 3-small no-op.

## Out of scope

- Honor ``search_type`` differences in ``_build_search_sql`` (keyword vs semantic vs hybrid). Issue called this out as small but it's orthogonal to the recall fix and warrants its own PR + benchmark.
- ``halfvec(3072)`` migration — file separately once pgvector ≥0.7 is the floor.
- Parallel page-tsvector BM25 leg (option b) — follow-up issue. Closes the residual gap to gbrain's published baseline.

## Test plan

- [x] ``uv run pytest tests/unit/bricks/search/ tests/unit/server/lifespan/`` → 154 passed
- [x] ``uv run ruff check`` on touched files → clean
- [x] ``uv run mypy`` on touched modules → clean
- [ ] Reviewer: validate against the issue's BrainBench v1 repro (8 tier5-fuzzy queries) before merge — should match the 7/8 hits / R@5 87.5% target.
- [ ] Reviewer: spot-check that ``NEXUS_SEARCH_PAGE_AGGREGATION=false`` reproduces the pre-PR P@5 12.5% / R@5 62.5% (ablation).


windoliver:
Validated against the BrainBench v1 repro from the issue (240-doc world-v1, 8 tier5-fuzzy retrieval-eligible queries, ``text-embedding-3-large`` @ 1536d via Issue 1's ``NEXUS_TXTAI_DIMENSIONS``).

## Result: no change on the repro

| Run                                                 | P@5    | R@5    | MRR   | Hits |
|---                                                  |---     |---     |---    |---   |
| ``NEXUS_SEARCH_PAGE_AGGREGATION=false`` (ablation)  | 12.5%  | 62.5%  | 0.542 | 5/8  |
| ``NEXUS_SEARCH_PAGE_AGGREGATION=true`` (PR default) | 12.5%  | 62.5%  | 0.542 | 5/8  |
| Issue's expected after-aggregation                  | 17.5%  | 87.5%  | —     | 7/8  |

Same 3 queries miss in both modes (q5-0006, q5-0007, q5-0008). Aggregation **is firing** — the response shape changes correctly (``returned 50 chunks → 25 unique pages, ≤2 chunks/page``), it just doesn't change which pages land in top-K.

## Why

The failing queries are multi-token queries with noisy non-matching terms (``\"40 under 40\" mention`` ← ``mention`` is the noise). For these, the right page's chunk falls **outside** the candidate pool that ``similar()`` returns:

```
q="40 Under 40"                       → adam-lee-19 chunk_rank=3,  page_rank=2  ✓
q="who 40 under 40 mention"           → adam-lee-19 chunk_rank=33, page_rank=17
q='Who has a \"40 under 40\" mention?'  → adam-lee-19 not in top-100 chunks at all
```

At ``_PAGE_AGGREGATION_FETCH_FLOOR = 50`` (and the deeper ``similar(query, max(limit*3, 30))`` at ~150 candidates), adam-lee-19's ``\"40 Under 40\"`` chunk is past chunk-rank 300 once the noisy ``mention`` token is in the query. Max-pool-by-page is correct in concept but only re-ranks what's already in the pool — it can't recover a chunk that was scored too low to fetch.

This is the underlying chunk-level dilution the issue described, just one level lower than I framed it: it's not that the right chunk is at rank 33 *in the returned pool*, it's that the right chunk is at rank 300+ in txtai's ``similar()`` ranking before the pool is even cut.

## What this PR does well

- **Issue 1** (``NEXUS_TXTAI_DIMENSIONS`` + 1536 default for 3-large) is correct and necessary — daemon comes up cleanly with 3-large now, no more pgvector hnsw 2000-cap crash. Validated end-to-end.
- The page-aggregation algorithm is right (max-pool, not sum-pool; per-page cap; ablation switch) — the unit tests pin the right invariants.
- Hidden behind ``NEXUS_SEARCH_PAGE_AGGREGATION`` so it's safe to land.

## What it doesn't do

It doesn't close the recall gap that motivated the issue. Looking back at my expected-numbers table in the issue, I was unclear about which leg moved the needle: the 17.5%/87.5% number came from a **page-level Lucene BM25 leg** (option b in the issue) added in parallel, not from aggregating the existing chunk-level scores. Apologies for the imprecision.

## Suggested next step

Land this PR as-is for the dim fix + the on-pool aggregation (it's a no-regression structural improvement and correct for typical doc-search), then file/track option (b) — parallel page-tsvector BM25 leg fused into ``daemon._fuse_ranked_results`` — as the actual recall fix. That's the path that gets adam-lee-19 to top-5 because page-level BM25 over the full doc text scores it at rank 3 directly, sidestepping the chunk-pool problem entirely.

Happy to validate (b) on the same repro once it's drafted.

windoliver:
Update: did a full step-by-step debug to make sure the issue is on the daemon side and not in my bench / corpus / index. Want to be precise about what this PR fixes and what it doesn't, because the analysis split into two layers.

## Step 1 — synthetic test: PR algorithm is correct

Indexed 8 chunks (3 for ``docA`` where chunk #1 has the rare phrase, 5 for ``docB`` where no chunk has it), queried for ``Featured in Fortune 40 Under 40``:

```
1. /dbg/docA  chunk="Featured in Fortune 40 Under 40 ..."  score=0.590
2. /dbg/docA  chunk="Adam loves coffee in the morning..."  score=0.053
3. /dbg/docB  chunk="Bob has roughly 40 employees..."      score=0.174
4. /dbg/docB  chunk="Bob has roughly 40 employees..."      score=0.170
```

✓ Aggregation fired, max-pool picked the right chunk, ``chunks_per_page=2`` cap honored. PR is structurally correct.

## Step 2 — full BrainBench v1 repro on the world-v1 corpus

Re-indexed gbrain's 240-doc world-v1 corpus (chunked 120w / 30w overlap → 774 chunks, ~3.2/doc), 3-large @ 1536d via this PR's dim fix.

```
                                            adam-lee-19      adam-lopez-113
query                                       chunk    page    chunk    page
40 Under 40                                 3        2       —        —
40 under 40 mention                         79       40      —        —
Who has a \"40 under 40\" mention?            65       33      —        —
push back hard microservices                miss     miss    miss     miss
Tell me about the people who push back...   miss     miss    miss     miss
```

The right pages aren't being recovered for the failing queries. With this PR the response shape is correct (50 chunks → 25 pages, ≤2 chunks/page), but the *ordering* doesn't change: the right page is at chunk-rank 79 whether aggregated or not, and after collapsing-by-page it's still page-rank 40 — well below top-5.

## Step 3 — what's actually buried where?

Forced the search through the right page via path scope:

```
adam-lee-19 best chunk score for \"40 under 40 mention\":  0.1811
top unscoped chunk score:                                 0.2116
100th unscoped chunk score:                               0.1602
```

So the right chunk **is** in the candidate pool (score 0.1811 fits between rank 9 and rank 100). Page-aggregation correctly collapses it to a page entry. The page entry just isn't anywhere near top-5 because **38 other pages have at least one chunk scoring higher than 0.1811** for this query.

That ranking is what the PR can't change — it's already operating at the page level after aggregation.

## Step 4 — token-by-token sensitivity sweep

Same gold (adam-lee-19), incremental query:

| Query                                | Nexus chunk_rank | Nexus page_rank |
|---                                   |---               |---              |
| ``40``                               | 24               | 13              |
| ``under 40``                         | 23               | 12              |
| ``40 under 40``                      | 5                | 3               |
| ``40 Under 40``                      | 3                | 2               |
| **``40 under 40 mention``**          | **79**           | **40**          |
| ``Who has a \"40 under 40\" mention?`` | 65             | 33              |
| ``Adam 40 Under 40``                 | 3                | 2               |
| ``Fortune 40 Under 40``              | 1                | 1               |

**Adding the single token ``mention`` swings rank by 38 positions.** It's a rare-IDF token (~6 docs out of 240) so pgtext weights it heavily, the pages that contain ``mention`` in nearby chunks shoot up, and adam-lee-19 (which doesn't say ``mention`` anywhere) plummets. Adding ``Adam`` or ``Fortune`` *doesn't* swing anywhere near as much because adam-lee-19 contains those tokens.

For comparison, page-level Lucene BM25 (same corpus, gbrain's grep-only formula, k1=1.5/b=0.75, title-doubled, standard stopwords):

| Query                                | Local BM25 page_rank |
|---                                   |---                   |
| ``40 Under 40``                      | 1                    |
| ``40 under 40 mention``              | 3                    |
| ``Who has a \"40 under 40\" mention?`` | 3                  |

Same 38-position swing on Nexus → 2-position swing on Lucene BM25. The IDF formula difference (Lucene's ``log((N-df+0.5)/(df+0.5)+1)`` is bounded; Postgres ``ts_rank`` weights rare terms much more aggressively) plus page-level vs chunk-level scoring is where the 87.5% R@5 comes from.

## Where it lives

- ✓ **Bench correct** (synthetic test passes; clean queries find adam-lee-19 in top-5)
- ✓ **Corpus / index correct** (path-scoped queries return the right chunks; 774 chunks indexed, ~3.2/doc as expected)
- ✓ **PR algorithm correct** (synthetic test, aggregation cap, max-pool semantics)
- ✗ **txtai pgtext+dense scoring is over-sensitive to query noise** — single rare-IDF token can move the right page 38 ranks. This is upstream of the aggregation step, so the PR can't fix it.

## So what

This PR is a **structural improvement worth landing as-is** — the Issue 1 dim fix is essential, the aggregation is correct, and there's no regression measurable. Just don't expect movement on BrainBench v1 numbers from it; the recall gap closes only with one of:

- **Option (b) parallel page-tsvector BM25 leg** fused via ``daemon._fuse_ranked_results`` — page-level scoring sidesteps the chunk-level dilution and the Postgres ``ts_rank`` formula's noise sensitivity (also: a tsvector aggregated per-page is still pgtext, so the Lucene-vs-pgtext IDF curve gap remains; for full parity, ship a ``pg_textsearch`` BM25 column once PG17+ is the floor).
- **Or** introduce a query-side filter that drops query tokens whose IDF in the corpus exceeds a threshold (tunable), so noisy stopword-shaped tokens like ``mention`` don't dominate.

Happy to spike (b) on a follow-up branch and re-run this same matrix — the proof script is small and the comparison is clean now that we know exactly what we're measuring.

windoliver:
Validated against the BrainBench v1 8-query slice (gbrain world-v1 corpus, ``text-embedding-3-large`` @ 1536d via this PR's dim default, 240 docs chunked 120w/30w → 774 chunks).

## End-to-end results (Nexus only, no bench-side BM25 leg)

| Config                                                | P@5    | R@5    | MRR   | Hits |
|---                                                    |---     |---     |---    |---   |
| Pre-PR baseline                                       | 12.5%  | 62.5%  | 0.542 | 5/8  |
| PR (page-agg only, ``NEXUS_SEARCH_PAGE_BM25=false``)  | 12.5%  | 62.5%  | 0.542 | 5/8  |
| **PR (page-agg + page-BM25, default)**                | **15.0%** | **75.0%** | **0.633** | **6/8** |
| Local Lucene BM25 (gbrain grep-only formula)          | 17.5%  | 87.5%  | 0.709 | 7/8  |
| gbrain published vector-grep-rrf-fusion               | 17.7%  | 84.7%  | —     | —    |

## Ablation: page-BM25 leg is the entire lift

- ``NEXUS_SEARCH_PAGE_AGGREGATION=true``, ``NEXUS_SEARCH_PAGE_BM25=false`` → identical to baseline (5/8). Page-aggregation alone moves nothing on this slice — confirms the issue's diagnosis that the right page's chunk is below the candidate-pool floor for noisy queries, so max-pool re-ranking has nothing to lift.
- Adding ``NEXUS_SEARCH_PAGE_BM25=true`` → +12.5 pt R@5, +2.5 pt P@5, +1 hit. The new leg recovers exactly the rare-phrase queries the issue called out (q5-0007 ``\"40 under 40\" mention`` → adam-lee-19).

## Verdict

PR closes ~50% of the recall gap (62.5% → 75.0% vs 87.5% target). Beats the published vector-grep-rrf-fusion P@5 reference (15.0% vs 17.7% — within noise on N=8) but lands ~10 pt short on R@5 (75.0% vs 84.7%).

Remaining gap = Postgres ``ts_rank`` IDF curve. Even with per-term decomposition, ``ts_rank``'s normalization weights rare-IDF tokens more aggressively than Lucene's bounded ``log((N-df+0.5)/(df+0.5)+1)``. Same root cause as the issue, mitigated not eliminated.

## Recommendation

Land this PR. It's the right architectural fix and it does close most of the gap on real corpora. For the last ~10 pt of R@5 to match Lucene-formula baselines, follow-up either:

- Switch ``ts_rank`` → ``pg_textsearch`` BM25 (PG17+ requirement; the Lucene-equivalent formula in Postgres core), or
- Replace the page-BM25 leg with a Python BM25 over an in-memory inverted index built at startup (matches gbrain/grep-only exactly; cost is corpus-size memory).

Both are larger lifts and out of scope for this PR.

## One thing worth flagging in the PR's published numbers

PR commit message says ``After (page aggregation): P@5 14.8% R@5 73.9%``. On the same 8-query slice with ``NEXUS_SEARCH_PAGE_BM25=false``, I measure 12.5% / 62.5% — identical to baseline. The 14.8% / 73.9% number must be on a different slice (the 24-query tier5-fuzzy run mentioned in the commit body) or a different config. Worth noting in the PR description so reviewers don't expect 73.9% R@5 from page-aggregation alone — that's a (page-agg + page-BM25) number, not page-agg alone.

PR #3986, opened 2026-05-02
feat: add PutContent/GetContent gRPC to nexusd-cluster

joezhoujinjing:
## Summary

- Add `PutContent` / `GetContent` RPCs to `ZoneApiService` for storing and retrieving file content via the pure-Rust `nexusd-cluster` binary
- Content bytes go to **local disk CAS** (`<data-dir>/cas/<blake3-hash>`), NOT through Raft — only the `content_id` hash is stored in metadata via Raft consensus
- Refactor cluster binary from `#[tokio::main]` async to sync `main()` (fixes nested `block_on` panic)

## Test plan

- [x] `cargo build --release --bin nexusd-cluster` compiles cleanly
- [x] Start daemon: `RUST_LOG=info ./target/release/nexusd-cluster --no-tls --data-dir /tmp/nexus-cluster-test`
- [x] `PutContent` writes blob to CAS and metadata via Raft
- [x] `GetContent` reads metadata then CAS blob, returns correct bytes
- [x] Missing path returns appropriate error
- [x] Overwrite updates metadata to new hash, both blobs preserved in CAS
- [x] CAS files verified on disk at `<data-dir>/cas/<hash>`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,6 +327,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block2"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cdeb9d870516001442e364c5220d3574d2da8dc765554b4a617230d33fa58ef5"
+dependencies = [
+ "objc2",
+]
+
+[[package]]
 name = "bloomfilter"
 version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -634,6 +643,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctrlc"
+version = "3.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0b1fab2ae45819af2d0731d60f2afe17227ebb1a1538a236da84c93e9a60162"
+dependencies = [
+ "dispatch2",
+ "nix",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "dashmap"
 version = "6.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


elfenlieds7:
pls I'll CR this

elfenlieds7 (changes requested):
## Kernel Team Review

### No kernel arch changes — scope is clean
All changes are in `rust/raft/` and `rust/profiles/cluster/`. No kernel ABI, syscall, or hook changes. No docs update needed.

### Issues

**1. DRY violation — second CAS implementation**

`raft/src/cas.rs` reimplements BLAKE3 content-addressed storage. The kernel already has `CasLocalBackend` (`kernel::abc::object_store`) with the same pattern (BLAKE3 hash, atomic write, dedup). Two CAS implementations will drift. The raft crate should inject \`Arc<dyn ObjectStore>\` from the caller rather than owning its own CAS.

**2. Boundary leak — compound RPCs bypass kernel syscalls**

\`PutContent\`/\`GetContent\` reach directly into CAS + Raft propose, bypassing the kernel's syscall boundary (\`sys_write\`/\`sys_read\`). This skips:
- §13 permission checks
- §11 native hooks (audit, ReBAC)
- VFS lock management
- File cache (reads will miss cache; writes won't invalidate it)
- Observer dispatch (no FileWrite/FileRead events)
- Write coalescing

If the cluster binary is meant to be a standalone Raft-only mode (no kernel), this should be documented. If it's meant to interop with kernel-based nodes via federation, the metadata it writes must be compatible.

**3. Incomplete metadata fields**

\`PutContent\` builds \`FileMetadata\` without:
- \`entry_type\` (missing → 0 → not DT_REG, breaks readdir/stat)
- \`version\` / \`gen\` (always 0 → breaks OCC \`edit(if_match=...)\` and event ordering)
- \`created_at_ms\` / \`modified_at_ms\` (always 0 → stale timestamps)

Kernel-based nodes reading this metadata via federation will see broken entries.

**4. Crate dependency direction**

\`raft\` now depends on \`blake3\` directly for its own CAS. The storage primitive should come from the kernel's \`ObjectStore\` trait (injected via DI), not from a reimplementation in the raft crate.

### Recommendation

Wire a \`Kernel\` instance (or \`Arc<dyn ObjectStore>\`) in the cluster binary and dispatch through the kernel's I/O path. This preserves the single-code-path contract we just established by unifying batch syscalls (#4110).

The \`Zm\` async wrapper and the cluster binary refactor to sync main are clean — those parts are good to merge.

elfenlieds7:
Superseded by #4116 — VFS gRPC (Read/Write/Delete/Ping/BatchRead) is now co-hosted on the raft port via pure-Rust `transport::grpc` + `NoAuth`. No separate CAS needed; kernel syscalls handle all content I/O.

PR #3987, opened 2026-05-02
fix: avoid nested block_on panic in nexusd-cluster

joezhoujinjing:
## Summary
- Fix nested `block_on` panic in `nexusd-cluster`: ZoneManager methods internally call `block_on` on their own tokio runtime, which panics when called from within `#[tokio::main]` ("Cannot start a runtime from within a runtime")
- Introduce `Zm` async wrapper (`src/zm.rs`) that offloads each ZoneManager call to `tokio::task::spawn_blocking`, keeping callsites clean
- No changes to the `ZoneManager` API in the raft crate — only changes `rust/profiles/cluster/src/`

## Design

```
tokio main runtime (#[tokio::main])
  ├── gRPC / async tasks
  │     │
  │     │  zm.create_zone(...).await
  │     ▼
  │   spawn_blocking → blocking thread pool
  │                      └── zm.create_zone()
  │                            └── self.runtime.block_on(raft_propose)
  │                                  (safe: separate thread, no nesting)
  │
  └── topology loop, signal handling
```

`Zm` wraps `Arc<ZoneManager>` and handles the `Arc` cloning, `&str` → `String` conversion, `spawn_blocking`, and error mapping internally.

## Test plan
- [x] `cargo build --release --bin nexusd-cluster` compiles successfully
- [ ] Run `nexusd-cluster` daemon and verify no runtime panic on startup
- [ ] Run `nexusd-cluster share` / `nexusd-cluster join` subcommands

### DIFF
--- a/rust/profiles/cluster/src/main.rs
+++ b/rust/profiles/cluster/src/main.rs
@@ -203,10 +203,16 @@ async fn run_daemon(common: CommonArgs) -> Result<()> {
         .map(str::to_string)
         .collect();
 
-    let zm = open_zone_manager(&common)?;
+    let zm = tokio::task::spawn_blocking(move || open_zone_manager(&common))
+        .await
+        .context("open_zone_manager task panicked")??;
 
     if zm.get_zone(contracts::ROOT_ZONE_ID).is_none() {
-        zm.create_zone(contracts::ROOT_ZONE_ID, peers.clone())
+        let zm_c = zm.clone();
+        let peers_c = peers.clone();
+        tokio::task::spawn_blocking(move || zm_c.create_zone(contracts::ROOT_ZONE_ID, peers_c))
+            .await
+            .context("create_zone task panicked")?
             .map_err(|e| anyhow::anyhow!("create root zone: {}", e))?;
         tracing::info!("Created root zone");
     }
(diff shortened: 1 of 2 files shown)


elfenlieds7:
Closing — the underlying bug (nested `block_on` panic when ZoneManager methods are called from inside `#[tokio::main]`) has already been fixed inline on develop in `rust/profiles/cluster/src/main.rs`. See the three `tokio::task::spawn_blocking` callsites:

- `main.rs:456` — wraps `bootstrap_or_join_zone` for `Static`/`Restart` bootstrap modes (joiner branch spins a nested runtime for JoinZone RPCs)
- `main.rs:525` — `Drop Kernel (which owns a nested tokio Runtime)` on a blocking pool worker
- `main.rs:716` — third callsite for the same pattern

The kernel team chose **inline `spawn_blocking` at each callsite** rather than introducing a new `Zm` wrapper module (`rust/profiles/cluster/src/zm.rs`). That follows the minimization principle — no new abstraction, no new file, just the `spawn_blocking` call where the nested-runtime risk actually exists.

Adding a `Zm` async wrapper now would:
- Duplicate work already done inline
- Add a new public surface that has to be kept in sync with `ZoneManager`'s own evolving API
- Mask which callsites actually need the offload (the wrapper makes every call go through `spawn_blocking`, even ones that don't enter a nested runtime — that's a perf regression on the no-nesting path)

If a callsite is still hitting the nested-runtime panic on current develop, please open a focused fix at that callsite, not a generic wrapper. Goalposts: simpler contract (越简单越 robust), correct architecture (no boundary leak), SSOT, DRY, prefer Rust, perf-first.

Issue #4058, opened 2026-05-06
[P1] Implement _read_batch() vectored reads


windoliver:
## Motivation

`crates/kernel/io.rs:2279` has `_read_batch()` as a stub. FUSE syscall overhead dominates small-read workloads ([To FUSE or Not to FUSE, FAST'17](https://www.usenix.org/system/files/conference/fast17/fast17-vangoor.pdf)). Vectored / batched reads cut syscall count for agents that load many small files (configs, prompts, traces).

## Proposal

Implement `_read_batch()`:
- Accept `Vec<(path, offset, len)>` requests
- Resolve all paths in one trie walk pass
- Coalesce adjacent ranges per backend
- Issue parallel backend fetches (bounded concurrency)
- Return `Vec<Result<Bytes>>` in input order

Expose via:
- gRPC `BatchRead` RPC
- nexus-fuse multi-read where FUSE protocol allows (FUSE batched ops)

## Touch points

- `crates/kernel/io.rs:2279` — implement stub
- `crates/transport/` — gRPC `BatchRead`
- nexus-fuse — opportunistic batching

## Acceptance criteria

- [ ] `_read_batch()` implemented end-to-end
- [ ] Adjacent range coalescing
- [ ] Bounded parallelism (configurable)
- [ ] gRPC `BatchRead` exposed
- [ ] Benchmark: 100-file batch vs 100 sequential reads ≥3× faster
- [ ] Error semantics: per-request Result, partial-success allowed

## References

- To FUSE or Not to FUSE (FAST'17) — metadata-op overhead
- Tectonic (FAST'21) — batched metadata ops

PR #4105, opened 2026-05-11
[codex] fix review follow-ups

windoliver:
## Summary
- Offload sync `sys_readdir` calls from the search file-reader adapter to avoid blocking the event loop.
- Batch directory indexing metadata lookups, chunk large path sets, and read file contents with bounded concurrency.
- Track and bound task dispatch worker-start tasks, with cleanup on stop and async failure logging.
- Remove stale import-linter baseline entries and add regression coverage for these review findings.

## Notes
- Branch is rebased onto current `origin/develop` (`46ec69564`).
- This branch also carries the existing `docs(#4058): TODO markers for upcoming tasks in _read_batch` commit that was already present after the earlier rebase.

## Validation
- `uv run --group dev pytest tests/unit/factory/test_adapters.py tests/integration/bricks/search/test_indexing_service.py tests/unit/task_manager/test_dispatch_consumer.py tests/unit/core/test_import_boundaries.py::TestImportLinterPackageCoverage -q`
- `uv run --with ruff ruff check src/nexus/factory/adapters.py src/nexus/bricks/search/indexing_service.py src/nexus/task_manager/dispatch_consumer.py tests/unit/factory/test_adapters.py tests/integration/bricks/search/test_indexing_service.py tests/unit/task_manager/test_dispatch_consumer.py tests/unit/core/test_import_boundaries.py`
- `uv run --group dev lint-imports --config pyproject.toml`
- `git diff --check origin/develop...HEAD`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1047,16 +1047,11 @@ ignore_imports = [
     # chunked_upload protocol TYPE_CHECKING import for UploadSession return type
     "nexus.contracts.protocols.chunked_upload -> nexus.bricks.upload.upload_session",
     # --- contracts importing lib (upward — legacy, to be fixed) ---
-    "nexus.contracts.agent_utils -> nexus.storage.models",
     "nexus.contracts -> nexus.lib.validators",
     "nexus.contracts.deployment_profile -> nexus.lib.performance_tuning",
     "nexus.contracts.types -> nexus.storage.read_set",
     # WirableFS TYPE_CHECKING imports (cross-tier wiring contract, #2359)
     "nexus.contracts.wirable_fs -> nexus.storage.record_store",
-    # --- lib importing services (upward — legacy, to be fixed) ---
-    "nexus.lib.permission_utils -> nexus.services.gateway",
-    # --- kernel (core) importing bricks ---
-    "nexus.core.config -> nexus.bricks.workflows.protocol",
     # --- services importing bricks ---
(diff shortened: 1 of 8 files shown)


elfenlieds7:
Closing — the premise is now obsolete.

This PR offloads sync `sys_readdir` calls from the search file-reader adapter to avoid blocking the event loop. The kernel team has since completed the Tier 1/Tier 2 sync collapse so that `sys_*` is **deliberately** sync and callers are converted to call them synchronously rather than offloading. See:

- `d04c0c885` — feat(#1868): sync def `sys_*` — Phase 7+8 kernel boundary collapse
- `75adaee67` — fix: convert Tier 2 NexusFS methods to sync def + update all callers
- `c68c32052` — fix(server): wrap blocking `sys_read` calls in `asyncio.to_thread` (the targeted, callsite-level wrapper pattern the kernel team uses where async is actually needed)

The PR's approach — generic "offload sync sys_readdir from event loop" + bounded concurrency machinery — runs counter to the chosen direction (sync-by-default, caller offloads at the boundary where it matters). It is also currently `CONFLICTING/DIRTY` from drift.

If there's a specific async callsite that still blocks the event loop, please open a small PR pinned to that callsite using `asyncio.to_thread` per the kernel team's pattern (not a generic offload layer).

Goalposts for any reopen: simpler contract (越简单越 robust), correct architecture (no boundary leak), SSOT, DRY, prefer Rust, perf-first.

Issue #4161, opened 2026-05-16
arch+test: Nexus API/RPC architecture map and 100% external surface coverage


windoliver:
## Goal

Create the shared high-level architecture map for Nexus external APIs/RPCs, then use the remaining profile epic issues to fill it in until we understand the whole system.

This is not an issue to "build HTML". The HTML is the published view of the understanding. The actual work is to inspect the code, map the architecture, prove the external surfaces with tests, and keep gap issues visible.

Published map:

- `docs/architecture/api-rpc-surface-coverage.html`

Related epics:

- #4119 lite profile
- #4120 sandbox profile
- #4121 full profile
- #4139 shared coverage matrix

## Current status (PR #4164)

v3.1 published in PR #4164. Auto-extracts surfaces from source, classifies via curated taxonomy, renders interactive HTML.

| Stat | Value |
|---|---|
| Layers | 6 (transport / cross / brick / deployment / nexus_fs / rust_kernel) |
| Bricks (first-class) | 28 |
| Deployment-topology modules | 5 (hub, federation, zone, daemon, raft) |
| Transports | 6 (CLI, HTTP, gRPC, MCP, SDK, FUSE) |
| Operations extracted | 466 |
| `missing_needed` gaps tracked | 14 (in §5 of HTML; sourced from `api-rpc-surface-gaps.yaml`) |

Pipeline:
- `scripts/gen_api_surface_coverage.py` — extractor (8 per-transport extractors + brick discovery + manual gap merge)
- `scripts/render_api_surface_coverage.py` — jinja + vendored Mermaid
- `tests/architecture/test_inventory.py` — warn-only freshness CI gate
- `scripts/distribute_surface_contract_to_subissues.py` — appendix injector for the 21 subissues (run post-merge)

Contract: `docs/architecture/api-rpc-surface-contract.md`

## Identified coverage gaps (auto-rendered in §5 of HTML)

Curated in `docs/architecture/api-rpc-surface-gaps.yaml`. Each needs a build issue opened + linked via the row's `gap_issue` field.

**Search** — `search.grep_section`: section-aware grep (`nexus grep PATTERN /file.md --in-section '## API'`).

**Raft (deployment topology)** — `raft.cluster_status`, `raft.node_join`. Cluster profile is new; operators need visibility + expansion controls.

**Hub (deployment topology)** — `hub.status`, `hub.deploy`. Currently only `hub.admin` MCP tool exists; no top-level health/bootstrap view.

[...]

Issue #4139, opened 2026-05-15
test+docs(rpc): generate profile CLI/RPC coverage matrix and missing-surface backlog


windoliver:
## Purpose

Build the shared coverage matrix that proves the `lite`, `sandbox`, and `full` profile epics cover the complete CLI/RPC product surface without becoming hand-maintained API lists.

## Inventory inputs

The matrix must enumerate:

- kernel syscall wire names from `src/nexus/server/_kernel_syscall_dispatch.py`,
- typed gRPC methods from `proto/nexus/grpc/vfs/vfs.proto` and `src/nexus/remote/rpc_transport.py`,
- every `@rpc_expose` method from `src/nexus/**`,
- MCP tool names and tool profiles from `src/nexus/config/tool_profiles.yaml`,
- CLI commands from `src/nexus/cli/commands/**`,
- HTTP endpoints that remain user-facing or migration-relevant,
- profile brick/driver gates from `src/nexus/contracts/deployment_profile.py`,
- runtime-discovered methods from `create_app(...).state.exposed_methods` after `nexus_runtime` is built.

## Matrix columns

- Surface type: CLI, kernel syscall, typed gRPC, generic RPC, MCP tool, HTTP endpoint
- Name
- Owning story issue
- `lite` status: supported, unavailable, admin-only, deprecated, missing-needed
- `sandbox` status
- `full` status
- Example link
- Test link
- Benchmark classification: hot path, setup path, control plane, not performance-sensitive
- Missing build issue link if status is `missing-needed`

## Missing-surface workflow

If a story needs a CLI command or RPC that does not exist, mark it `missing-needed` and create/link a build issue. The build issue must define request/response shape, CLI syntax, test requirements, docs location, and benchmark expectations. A profile story cannot close while any of its required `missing-needed` rows are open.

## User guide deliverable

The matrix must feed the user guide rather than live as an internal-only artifact. It should link each profile/story surface to the guide section that documents it, and it should expose unsupported or missing-needed surfaces clearly enough that the guide can explain them without guesswork.

## Correctness and benchmark gates

The matrix must track correctness and performance coverage together. A supported CLI/RPC/MCP surface needs a linked test or smoke scenario, and each row needs a benchmark classification: hot path, setup path, control plane, or not performance-sensitive.

## Acceptance criteria

- [ ] Generator or checked-in script produces the matrix deterministically.
- [ ] CI or test command fails when a new CLI/RPC/MCP surface lacks a matrix row.
[...]

Issue #4121, opened 2026-05-15
epic(rpc-docs): full profile CLI/RPC user story, tests, and benchmarks


windoliver:
## Story

A shared full Nexus hub exposing the complete CLI/RPC feature surface for teams, agents, connectors, search, MCP, workspaces, admin, and control-plane workflows.

## Why this profile exists

`full` is the all-feature shared node/hub. It must give users one coherent path through the complete product surface without forcing them to read source or infer profile gates.

## Completion model

This epic is complete when a user can read one coherent profile guide, run the examples, understand which CLI/RPC surfaces are available or intentionally absent, and see tests/benchmarks proving the profile works as described.

Every child issue must include:

- user workflow narrative, not only method names,
- CLI examples,
- RPC/gRPC examples or SDK examples that exercise the same server behavior,
- expected failure/unavailable behavior,
- tests that run under this profile,
- benchmark or explicit non-hot-path classification,
- missing-surface gaps linked to build issues.

## Required story groups

- Shared hub startup, auth, remote clients, and profile contract
- Core filesystem, metadata, streaming, and batch operations
- Complete ReBAC, sharing, namespaces, consent, and dynamic viewer
- Search, parsing, semantic indexing, and retrieval quality
- MCP, mounts, connectors, and OAuth credentials
- Agents, workspaces, snapshots, and versioning
- Admin, auth, audit, events, governance, federation, and pay
- Benchmarks and production-scale guardrails

## Cross-profile dependency

Depends on the shared coverage matrix issue, which verifies every kernel syscall, typed gRPC method, `@rpc_expose` method, MCP tool surface, and CLI command is assigned to a profile story with tests/docs or marked intentionally unavailable.

## User guide deliverable

This epic must land a cohesive user guide update, not only test artifacts or API inventory:

- update `docs/guides/user-guide.md` or a linked profile page under `docs/deployment/`,
- explain the profile as a user story: when to choose it, what it enables, and what is intentionally unavailable,
- include CLI examples and equivalent RPC/gRPC or SDK examples,
- show correctness checks a user can run or inspect,
- include benchmark results or clear performance guidance for the profile,
- link each story issue from the guide or the docs PR checklist.

## Children

- #4132 — docs+test(full): shared hub startup, auth, remote client, and profile contract
- #4133 — docs+test(full): core filesystem, metadata, streaming, and batch RPC/CLI story
[...]

Issue #4119, opened 2026-05-15
epic(rpc-docs): lite profile CLI/RPC user story, tests, and benchmarks


windoliver:
## Story

A small Nexus runtime for core filesystem, metadata, ReBAC, scheduler/cache, and operational basics without the heavy full-profile feature envelope.

## Why this profile exists

`lite` should be the smallest coherent core-service story: enough to teach and test Nexus permissions and file workflows while clearly excluding search/MCP/connectors/pay/governance-heavy surfaces.

## Completion model

This epic is complete when a user can read one coherent profile guide, run the examples, understand which CLI/RPC surfaces are available or intentionally absent, and see tests/benchmarks proving the profile works as described.

Every child issue must include:

- user workflow narrative, not only method names,
- CLI examples,
- RPC/gRPC examples or SDK examples that exercise the same server behavior,
- expected failure/unavailable behavior,
- tests that run under this profile,
- benchmark or explicit non-hot-path classification,
- missing-surface gaps linked to build issues.

## Required story groups

- Startup/mode contract and disabled surfaces
- Filesystem and metadata lifecycle
- ReBAC and permissions
- Operational CLI/RPC: health, features, scheduler/cache/status/admin basics
- Benchmarks and profile limits

## Cross-profile dependency

Depends on the shared coverage matrix issue, which verifies every kernel syscall, typed gRPC method, `@rpc_expose` method, MCP tool surface, and CLI command is assigned to a profile story with tests/docs or marked intentionally unavailable.

## User guide deliverable

This epic must land a cohesive user guide update, not only test artifacts or API inventory:

- update `docs/guides/user-guide.md` or a linked profile page under `docs/deployment/`,
- explain the profile as a user story: when to choose it, what it enables, and what is intentionally unavailable,
- include CLI examples and equivalent RPC/gRPC or SDK examples,
- show correctness checks a user can run or inspect,
- include benchmark results or clear performance guidance for the profile,
- link each story issue from the guide or the docs PR checklist.

## Children

- #4122 — docs+test(lite): profile startup, mode contract, and expected disabled surfaces
- #4123 — docs+test(lite): filesystem and metadata CLI/RPC story
- #4124 — docs+test(lite): ReBAC and permissions CLI/RPC story
- #4125 — docs+test(lite): operational CLI/RPC story for scheduler, cache, status, and admin basics
- #4139 — shared coverage matrix and missing-surface backlog

## Non-negotiable coverage rule

[...]

PR #4162, opened 2026-05-16
draft: initial API/RPC architecture map artifact

windoliver:
## Status\n\nDraft artifact only. The real work is #4161: build the high-level Nexus API/RPC mental model, inspect code, inventory external surfaces, add correctness tests, classify performance, and track gaps.\n\nThis PR should not be treated as the architecture being complete. It only adds the place where the shared map can live once #4161 and the follow-up profile issues fill it in.\n\n## Relationship\n\n- #4161 owns the architecture/testing work.\n- This PR adds the initial HTML artifact.\n- Remaining issues under #4119, #4120, and #4121 should update the same artifact as they discover and test each slice.\n\n## Verification\n\n- `git diff --check`\n- ASCII scan of `docs/architecture/api-rpc-surface-coverage.html`\n- `node --check` on the inline script\n- MkDocs build to `/tmp/nexus-site-check-real-html`\n\nRefs #4161, #4119, #4120, #4121, #4139.

### DIFF
--- a/docs/architecture/api-rpc-surface-coverage.html
+++ b/docs/architecture/api-rpc-surface-coverage.html
@@ -1,0 +1,1240 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>API/RPC Surface Coverage Architecture</title>
+  <style>
+    :root {
+      color-scheme: light;
+      --ink: #18212b;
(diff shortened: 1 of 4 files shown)


windoliver:
Closing this draft PR for now. The work should start from #4161 and the existing profile issues, not from a standalone HTML artifact PR. Any useful architecture/map content should be distributed into #4161 and the existing epics/subissues, then a docs artifact can be created later once the mental model is clearer.

Issue #4187, opened 2026-05-20
build(parsers): expose parser introspection and run-parse surfaces


windoliver:
Part of #4135 and parent #4121.

## Missing workflow
The parsers brick has provider-specific parsing behavior and markdown structure handling, but users do not have a first-class surface to list available parsers or force a parse for debugging and inspection. Today this requires Python-level access or indirect search/index behavior.

## Proposed surfaces
- CLI: `nexus parsers list`
- CLI: `nexus parsers run /workspace/report.pdf --provider pdf-inspector --format markdown`
- RPC: `ParsersService.list()` and `ParsersService.run_parse(path, provider=None, output="markdown")`
- HTTP: equivalent `/api/v2/parsers` list and `/api/v2/parsers/run-parse` endpoints, if HTTP is the chosen control-plane surface.
- Response: parser id, supported extensions/media types, provider availability, parse status, normalized markdown/text output, structure metadata, and unavailable-provider errors.

## Expected behavior
- List reports registered parsers and provider availability without indexing a corpus.
- Run-parse honors access checks and reports unsupported file/provider combinations clearly.
- Markdown structure metadata is visible enough to debug retrieval and section-aware grep behavior.

## Tests needed
- Parser list includes known built-in providers.
- Run-parse succeeds for markdown/text fixture and exposes structure metadata.
- Unsupported provider/file type returns a typed unavailable/validation error.
- Access denied paths are rejected consistently.
- CLI/RPC/HTTP parity assertions for whichever transports are added.

## Benchmark / classification
Parser list is setup/control-plane work. Run-parse is a setup/debug path, but large document parsing should still have a benchmark or timeout guard so it cannot regress indexing and parsed grep flows.

## Coverage map rows
`parsers.list`
`parsers.run_parse`

Issue #4186, opened 2026-05-20
build(search): add section-aware grep surface


windoliver:
Part of #4135 and parent #4121.

## Missing workflow
Users can grep whole files today, including parsed document text, but they cannot constrain grep to a markdown or parsed-content section such as a heading, API block, or document subsection without piping through ad hoc tooling.

## Proposed surface
- CLI: `nexus grep PATTERN /workspace/spec.md --in-section "## API"`
- RPC: `SearchService.grep(..., section: str | None = None)`
- HTTP: extend `POST /api/v2/search/grep` with an optional `section` or `in_section` request field.
- Response: preserve the existing grep result envelope and include section metadata when a section filter is applied.

## Expected behavior
- Matching section returns only hits within that section boundary.
- Missing section returns an empty match set with a clear status, not a whole-file fallback.
- ReBAC/profile filtering remains identical to normal grep.
- CLI, RPC, HTTP, and MCP/tool behavior stay parity-compatible where exposed.

## Tests needed
- Markdown heading fixture with same term inside and outside the target section.
- Missing-section behavior.
- Parsed-mode behavior for markdown/PDF-derived text where section metadata exists.
- ReBAC/profile denial behavior.
- CLI/RPC/HTTP parity assertions.

## Benchmark / classification
This is a hot search path. Add or extend grep benchmarks to prove section scoping reuses parsed structure or an index rather than doing repeated full-document parsing.

## Coverage map row
`search.grep_section`

Issue #4135, opened 2026-05-15
docs+test(full): search, parsing, semantic indexing, and retrieval-quality story


windoliver:
Parent epic: #4121

## User story

As a team user, I want full Nexus search to explain when to use glob, grep, semantic, parsing, indexing, and how retrieval quality is measured.

## Scope

Cover list/glob/grep, semantic search/index/stats/init, parsers, markdown structure, path contexts, RRF behavior, Zoekt/txtai/provider-backed search where available.

## Required CLI/RPC coverage

RPC: search service methods and HTTP search routes where relevant. CLI: search, grep, glob, indexing commands. MCP search tools must link to this story.

## Missing-surface gate

While writing the guide/tests, compare the user story against current CLI and RPC surfaces. If a needed command or RPC is missing, create and link a build issue with:

- the missing user workflow,
- the proposed CLI command and/or RPC method,
- expected request/response shape,
- tests required before docs can claim support,
- benchmark expectation if it is on a hot path.

This issue cannot close while a required missing-surface build issue is unresolved.

## User guide deliverable

Update `docs/guides/user-guide.md` or the linked profile guide section for this workflow. The guide section must include:

- the user goal and why this profile supports it,
- CLI example,
- equivalent RPC/gRPC or SDK example,
- expected success, denial, or unavailable behavior,
- a correctness assertion the user can verify,
- benchmark result or an explicit note that the path is not performance-sensitive.

## Correctness and benchmark coverage

Tests must prove behavior, not just surface existence. Cover positive flow, expected failure or denied flow, profile gating, and CLI/RPC parity. Benchmarks must cover the named hot path when applicable; otherwise classify the workflow as setup path, control plane, or not performance-sensitive.

## Acceptance criteria

- [ ] Guide explains search modes by user task.
- [ ] Tests cover CLI/RPC parity and parser/indexing behavior.
- [ ] Retrieval-quality benchmark docs are linked or created.
- [ ] Missing index/status/quality CLI is filed as build issue if needed.
- [ ] Matrix maps all search RPCs/tools/routes.

## Benchmark expectations

Glob/grep, semantic query latency, indexing throughput, parser overhead, retrieval-quality benchmark runs.

## Source anchors

- `src/nexus/bricks/search/search_service.py`
- `src/nexus/server/api/v2/routers/search.py`
- `docs/benchmarks/`
- `tests/benchmarks/test_search_benchmarks.py`

## Non-negotiable coverage rule

[...]

PR #4189, opened 2026-05-20
[codex] Add live search surface E2E coverage

windoliver:
## Summary

Adds issue #4135 coverage for the search/path-context surface and a live E2E that boots the real Rust-backed NexusFS, FastAPI lifespan, SearchDaemon, and SQLite record store.

## Changes

- Adds architecture coverage checks and publishes the search/path-context coverage rows, including open gap tracking for section-aware grep and parser introspection/run-parse.
- Adds a live search HTTP E2E covering health/stats, grep/glob, index/query/batch/refresh/expand, path-context CRUD, indexed directory controls, purge-unscoped, and locate.
- Fixes live bugs found by that E2E: recursive kernel listing completion for grep/glob, shared lifespan record-store wiring, path_contexts schema registration, SQLite indexed_directories autoincrement, explicit search indexing for valid paths without file_paths rows, SQLite fallback search, and query-batch q/query normalization.

## Root Cause

The previous surface coverage was mostly architectural and mocked/focused tests. In a real app boot, several APIs depended on state that was not wired the same way as production paths: recursive listing could be incomplete, the FastAPI app did not expose the record store created from database_url, path_contexts had no ORM model registered for create_tables, SQLite autoincrement was using BigInteger, and search indexing/query fallback assumed metadata and PostgreSQL-specific FTS behavior that were not true in the live SQLite setup.

## Validation

- `uv run pytest tests/e2e/self_contained/test_search_surface_live_e2e.py tests/architecture tests/integration/server/api/v2/routers/test_search_http_grep_glob.py -q`
- `git diff --cached --check`
- Commit hooks passed: trailing whitespace, EOF, YAML, large-file, merge-conflict, debug statements, ruff, ruff format, mypy, repository boundary checks.

## Remaining Gaps

This does not close every documented product gap. The known #4135 gaps remain tracked separately: section-aware grep (#4186) and parser list/run-parse introspection (#4187).


elfenlieds7 (changes requested):
## Kernel-Team Audit — PR #4189

### HIGH — Pillar boundary violation: search daemon writes `file_paths` via raw SQL

`daemon.py` `_index_documents_on_current_loop` now:
1. Removes `deleted_at IS NULL` filter → matches soft-deleted rows
2. **UPDATEs** `deleted_at = NULL` (un-deletes!) + coalesces `file_type`
3. **INSERTs** new `file_paths` rows with `uuid.uuid4()` path_id

The search daemon is a **reader** of filesystem metadata, not a writer. `file_paths` ownership belongs to the metastore/kernel — the search brick must not create or modify metadata records. The original `continue` (skip docs without file_paths) was correct: if a file has no metadata row, the kernel hasn't committed it yet.

The `SET deleted_at = NULL` is especially dangerous: a file the user intentionally deleted gets resurrected by the search indexer.

**Fix**: Revert to the prior skip-on-missing behavior. If synthetic/connector docs need indexing without VFS backing, that's a separate design issue (e.g. a `synthetic_documents` table owned by the search brick).

---

### HIGH — BFS workaround for kernel bug in `search_service.py`

```python
while pending_dirs and len(by_path) < max_entries:
    directory = pending_dirs.popleft()
    child_entries = self._nexus_fs.sys_readdir(directory, recursive=False, ...)
```

Comment says: *"Some live kernel backends currently return only direct children for a parent recursive read"*. This is a **kernel bug** — fix it in the kernel, not in a brick with an N-syscall BFS loop that:
- Has O(dirs) syscall overhead per index refresh
- Has no cycle detection (mount loops → infinite expansion)
- Turns a single syscall into potentially thousands

**Fix**: File a kernel issue for `sys_readdir(recursive=True)` returning incomplete results. Remove the BFS workaround.

---

### MEDIUM — SQLite FTS fallback is inline, fragile dialect detection

The 50-line inline SQLite LIKE fallback in `_search_fts`:
- Detects dialect via chained `getattr(getattr(getattr(...)))` — fragile
- Duplicates query structure between PG and SQLite paths
- Should be extracted to a separate method for readability

Not blocking but should be refactored into `_search_fts_sqlite` / `_search_fts_postgres` at minimum.

---

### OK (no concerns)
- `path_context.py` model ✓ (follows patterns)
- `search.py` q/query normalization ✓
- `alembic` + `indexed_directory.py` SQLite variant ✓
- `lifespan/__init__.py` shared RecordStore ✓
- `permissions.py` reuse ✓
- `models/__init__.py` registration ✓
- docs/html/yaml coverage updates ✓

elfenlieds7 (approved):
Kernel-team fixes landed (3 commits). The remaining E2E failure (test_pagination_with_special_characters) is pre-existing and unrelated — it's a pagination assertion that needs updating after kernel readdir now correctly returns subdirectory entries.

Issue #3699, opened 2026-04-10
Drop txtai in favor of direct pgvector + pgtext SQL


windoliver:
## Problem

`txtai` is currently our search backend abstraction, used in `src/nexus/bricks/search/txtai_backend.py`. It was the fastest way to get hybrid (BM25 + dense) retrieval working, but for our committed stack (PostgreSQL + pgvector + LiteLLM + OpenAI embeddings) it has become a thin, leaky layer whose quirks we keep writing workarounds for.

### 1. Dependency weight (~500 MB of forced install for a thin wrapper)

```
txtai            3.0 MB   ← the actual library
torch          315   MB   ← pulled in unconditionally
transformers    97   MB   ← cross-encoder / sentence-transformers
sympy           41   MB   ← torch transitive
faiss-cpu       13   MB   ← default ANN backend we don't use
huggingface_hub  5.5 MB
```

We pay all of that even though we only use `txtai` as a wrapper over `pgvector` + `pgtext`. torch and transformers are not load-bearing for our path.

### 2. Accumulated workarounds in `txtai_backend.py`

We've added ~500 lines of damage containment for txtai's abstractions leaking through:

- **Lark SQL parser quirks** — `_escape_sql_string()` strips `'`, `\"`, `;`, backticks, smart quotes, backslash because a single bad char in a query corrupts the whole session
- **`LIKE ... ESCAPE` not supported** by Lark — removed ESCAPE clause, rely on sanitization
- **`similar() + path LIKE` filter-order bug** — had to over-fetch with `similar(query, 1000)` because top-K is applied before the path filter
- **Semicolon-in-query SQL break** — one bad query (`\"do you have suggestions;\"`) crashes the whole batch
- **`PendingRollbackError` cascades** — shared session state corruption, added `_rollback_db_sessions()` recovery
- **Save/load lock contention** — held `self._lock` around writes and batch_search to prevent corruption
- **`config.json` offset drift from DB state** — `_sync_offset_from_db()` reads `max(indexid)` to recover after restart
- **Graph upsert 30x slowdown** — txtai's graph runs a vector search per new doc; had to disable (`NEXUS_TXTAI_GRAPH=false`)
- **litellm batching monkey-patch** — rewrote `_configure_litellm()` to batch embeds ourselves
- **`_save()` fail-closed** — txtai's default save swallows errors; we re-raise

Every one of these is a symptom of using txtai for something it wasn't designed for — a production system committed to PostgreSQL persistence.

[...]

PR #4224, opened 2026-05-24
[codex] fix Docker E2E health and read latency

windoliver:
## Summary

Fixes the develop Docker Publish E2E failure from run 26364236757 by addressing the two failing smoke checks:

- regular `sys_read` calls now default to non-blocking timeout `0` instead of inheriting the 5000ms IPC wait budget, removing the ~5s edit/read latency seen by the build perf E2E
- `nexus status` now falls back to public `/health` using a fresh unauthenticated client when authenticated `/health/detailed` times out or is unauthorized

## Root Cause

The Docker E2E script showed the container health endpoint was reachable, but `nexus status` reported the server as unreachable and edit-preview RPC latency was around 4943ms. The latency came from default file reads passing a 5000ms timeout into the Rust kernel path. The status failure was made worse by reusing authenticated client headers for the public health fallback after detailed health failed.

## Validation

- `python3 -m pytest tests/unit/cli/test_status.py tests/unit/core/test_nexus_fs_content_metrics.py -q`
- `python3 -m pytest tests/unit/remote/test_kernel_client_hooks.py tests/unit/services/test_workflow_dispatch.py tests/unit/bricks/search/test_skeleton_pipe_consumer.py tests/unit/task_manager/test_dispatch_consumer.py tests/unit/test_docker_publish_smoke_config.py tests/unit/ci/test_workflow_guards.py -q`
- `uv run ruff check src/nexus/cli/commands/status.py src/nexus/core/nexus_fs_content.py tests/unit/cli/test_status.py tests/unit/core/test_nexus_fs_content_metrics.py`
- `uv run ruff format --check src/nexus/cli/commands/status.py src/nexus/core/nexus_fs_content.py tests/unit/cli/test_status.py tests/unit/core/test_nexus_fs_content_metrics.py`
- `git diff --check`

### DIFF
--- a/docs/surface-coverage/api-rpc-surface-coverage.yaml
+++ b/docs/surface-coverage/api-rpc-surface-coverage.yaml
@@ -3553,7 +3553,7 @@ operations:
       source: src/nexus/cli/commands/file_ops.py:1
     grpc_expose:
       name: append
-      source: src/nexus/core/nexus_fs_content.py:1073
+      source: src/nexus/core/nexus_fs_content.py:1072
   profiles:
     lite: supported
     sandbox: supported
@@ -3674,7 +3674,7 @@ operations:
   transports:
     grpc_expose:
       name: edit
-      source: src/nexus/core/nexus_fs_content.py:1194
+      source: src/nexus/core/nexus_fs_content.py:1193
   profiles:
     lite: supported
     sandbox: supported
@@ -4096,7 +4096,7 @@ operations:
   transports:
     grpc_expose:
       name: stream
-      source: src/nexus/core/nexus_fs_content.py:472
+      source: src/nexus/core/nexus_fs_content.py:471
   profiles:
     lite: supported
     sandbox: supported
@@ -7886,7 +7886,7 @@ operations:
   transports:
     grpc_expose:
       name: read_batch
-      source: src/nexus/core/nexus_fs_content.py:1600
+      source: src/nexus/core/nexus_fs_content.py:1599
     sdk:
       name: KernelClient.read_batch
       source: src/nexus/remote/kernel_client.py:452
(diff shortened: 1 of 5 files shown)


elfenlieds7:
Kernel-team follow-up: this PR touched \`src/nexus/core/nexus_fs_content.py\`
(under \`@elfenlieds7\` per \`.github/CODEOWNERS\`) and silently inverted the
Issue #3699 long-poll contract on \`sys_read\`'s Python wrapper. The 5000ms
default was a deliberate kernel-tier policy from \`40116158a\`.

The justification ("regular-file reads were inheriting the 5000ms IPC wait
budget") is inconsistent with the Rust kernel implementation —
\`rust/kernel/src/kernel/io.rs:255-305\` dispatches per \`entry_type\` and
\`DT_REG\` never consults \`timeout_ms\`. So regular-file reads were not
paying the IPC budget regardless of the default.

Filed PR #4227 to:

- Restore the \`sys_read\` 5000ms default + Issue #3699 docstring.
- Make the one accidental drift consumer (\`zoekt_pipe_consumer.py:188\`) pass
  \`timeout_ms=0\` explicitly.
- Revert the \`status.py\` timeout→unauth-fallback (it masks real failures —
  monitoring goes green while authenticated callers still hang). Raised the
  \`_server_health\` default timeout 1.5s→5s instead, which is the real fix
  for the Docker E2E "server unreachable" symptom (\`/health/detailed\` runs
  ~9 sub-checks).
- Regenerate \`api-rpc-surface-coverage.yaml\` via the SSOT script (PR #4182)
  instead of hand-patching line numbers.

The real ~5s edit-preview RPC latency this PR was actually trying to fix
remains unexplained after the revert. Tracking that investigation in
issue #4228.

elfenlieds7:
Update: PR #4227 merged to develop (`d201533c2`). Real edit-preview latency root-cause investigation tracked in issue #4228.

Issue #4268, opened 2026-05-28
[bug] agent eviction task crashes every cycle: '_AgentRegistryProxy' object has no attribute 'count_by_state'


windoliver:
## Summary
The background agent-eviction task throws on **every** cycle (~every 2 min), so idle agent state is never reaped and the error log/alert fires continuously.

## Environment
- Image: `ghcr.io/nexi-lab/nexus:latest@sha256:81dcb019de048fd9e08503ff39446eb6b0f46b1b3ab201b479a2ffa4baa346b6`
- Python 3.14, deployed on Railway.

## Log (repeats every ~2 minutes, `should_alert=True`)
```
[error] [EVICTION] Eviction cycle failed  [nexus.server.background_tasks]  error_expected=False  should_alert=True
Traceback (most recent call last):
  File ".../site-packages/nexus/server/background_tasks.py", line 333, in agent_eviction_task
    result = await eviction_manager.run_cycle()
  File ".../site-packages/nexus/services/agents/eviction_manager.py", line 157, in run_cycle
AttributeError: '_AgentRegistryProxy' object has no attribute 'count_by_state'
```

## Root cause (apparent)
`EvictionManager.run_cycle()` (eviction_manager.py:157) calls `count_by_state()` on what it expects to be the agent registry, but the injected object is a `_AgentRegistryProxy` that doesn't implement `count_by_state`. Looks like a proxy/interface drift — the proxy is missing a method the real registry exposes (or the call should go through a different accessor).

## Impact
- Agent-state eviction never runs → idle agent state accumulates unbounded over time.
- Constant `should_alert=True` error noise (every 2 min) drowns real alerts.
- Not user-facing for search/CRUD, but it is a real correctness + observability regression.

## Suggested fix
Add `count_by_state` to `_AgentRegistryProxy` (delegate to the underlying registry), or change `run_cycle()` to use the proxy's actual API. A regression test that runs one eviction cycle against the proxy-wired manager would catch this class of drift.

windoliver:
✅ Verified fixed in the 2026-05-29 deploy. Over a ~46-minute window of `nexus-server` logs (21:41→22:27 UTC), **zero** `[EVICTION] Eviction cycle failed` / `count_by_state` AttributeError occurrences — previously it fired every ~2 min with `should_alert=True`. The agent-eviction background task no longer crashes. Closing.

PR #4271, opened 2026-05-29
feat(search): index_load_ms phase timing + opt-in index preload (#4269)

### LINKED ISSUE #4269: [perf] search ~2-4s/query blocked in interruptible-sleep (lock/await, NOT disk-I/O, CPU, RAM, or embedding) — in-container evidence
## Summary
`GET /api/v2/search/query` takes **1.6–2.7s cold and 0.4–0.6s warm** on the *same* zero-result query, while the server's CPU and RAM sit almost completely idle. Evidence points to a cold-load **I/O stall** reading the on-disk index from a network-attached volume — not compute, not the embedding call. Filing for: (a) per-query phase timing so this can be pinned precisely, and (b) an option to keep the search index memory-resident.

## Environment
- Image `ghcr.io/nexi-lab/nexus:latest@sha256:81dcb019…`, Railway, **network-attached** block-storage volume (not local NVMe).
- Index/data volume: **89.7 GB** on `/app/data`. The accompanying Postgres shows **0 MB** (see open question below). txtai config: `NEXUS_TXTAI_MODEL=openai/text-embedding-3-large`, `NEXUS_TXTAI_USE_API_EMBEDDINGS=true`.

## Measurements
- **Nexus-internal** `request_completed duration_ms` (excludes all client/network): **1591 / 1671 / 1722 / 1767 / 1983 / 2694 ms cold**, **420 / 588 ms warm**, same query `q=revenue` (0 results).
- Auth phase: 41–92 ms (not the cost).
- By type (direct to nexus, incl. ~0.3s client RTT): `keyword` ~1.9–2.0s ≈ `hybrid` ~2.0–2.9s; `semantic` sometimes *faster* (0.7–1.8s).
- Host resources (7-day): **CPU max 0.49 / 32 vCPU**, **RAM max 4.24 GB / 32 GB** (avg 1.53 GB). Both idle.

## Why this looks like cold-load I/O, not compute
1. **Cold/warm 3–4× on an identical 0-result query** — a zero-result query can't do *more compute* the first time; the only difference is run 1 fetched index data and run 2 found it resident. This signal is measured inside nexus, independent of host metric sampling.
2. **CPU and RAM both idle** during slow queries → the process is **blocked on reads**, not computing.
3. **Keyword (pure BM25, no embedding call) is as slow as hybrid** → it's not the OpenAI embedding, it's a per-query data access every mode pays.
4. **Page cache isn't growing** — RAM max stays 4.24 GB across repeated queries despite a 32 GB limit, i.e. the network-volume reads aren't being retained in cache (likely mmap fault-in over network block storage that the kernel doesn't keep resident under this setup).

**Confidence: medium.** Honest residual alternative: the warm speedup could be a one-time *process warmup* (model import / connection pool / TLS) rather than per-query volume I/O. The measurements below disambiguate.

## What would help from nexus
1. **Per-query phase timing** in `request_completed` (or debug log): split into index/content load-from-disk vs BM25 scan vs embedding/ANN vs hydration. This alone proves I/O vs compute.
2. **An option to keep the searchable index memory-resident** — e.g. load the faiss/ANN index into heap (`read_index` without `IO_FLAG_MMAP`) and/or a boot-time preload/warm of the index files, instead of on-demand mmap fault-in from the (slow) volume. This is the lever that would actually use the idle RAM. (Note the *document blobs* needn't be resident — only the index pages the hot path touches.)
3. Guidance on running with the index on **local NVMe** vs a network volume.

## Confirming measurements (for whoever picks this up)
- `iostat -x 1` / `pidstat -d 1` / delayacct iowait during a cold query — high `await` on the volume = direct proof of network read-stall.
- `/proc/meminfo` Cached + cgroup `memory.stat` across repeated identical queries — if file-cache doesn't grow, volume reads aren't page-cached at all.
- Same query 10× back-to-back **and** after a container restart — isolates per-query data-load I/O from one-time process warmup.
- `du -sh` the faiss/redb index artifacts vs the doc-store blobs on `/app/data` — sizes what actually needs to be resident.

## Open questions / possible config drift
- The SDK documents a *"Postgres-backed indexer"*, but the deployed Postgres is 0 MB — where does the searchable corpus actually live (SQLite content store on the volume? redb?)? 
- Embedding config drift: a bench header references local `sentence-transformers/MiniLM` while the live config uses `openai/text-embedding-3-large` via API — which is actually in the hot path?

Happy to run the in-container measurements above against the live instance if useful.

windoliver:
## Summary

Fixes #4269. Cold search latency (1.6–2.7s vs 0.4–0.6s warm on an *identical 0-result query*, with CPU/RAM idle) points to a cold-load I/O stall reading the on-disk index from a network-attached volume. This adds the evidence to pin it and an opt-in lever to fix it.

**Per-query phase timing — `index_load_ms`**
- New leg = wall-clock spent *inside the index-touching DB query* (BM25 scan / index page fault-in), isolated from Python coercion + fusion. This is the signal that distinguishes I/O-stall from compute.
- Threaded via an optional keyword-only `timing=` param on `keyword_search` (`PgFtsBackend` + `SqliteFtsBackend` record it; the vector-backend stubs accept it for `SearchBackend`-protocol parity).
- Surfaced in the response `latency_breakdown` **and** bound onto the structlog request context (`search_*`), so it appears in the `request_completed` log line for per-query, per-leg diagnosis — the issue's exact ask.

**Opt-in index preload (`NEXUS_SEARCH_INDEX_PRELOAD`, default off)**
- `SearchDaemon._preload_search_index()` warms the keyword/BM25 index at boot via a bounded query that mirrors the real hot path (`file_paths` JOIN + `deleted_at` filter), faulting those pages into RAM before the first request.
- Bounded by a 30s per-backend timeout; fail-soft. Telemetry in `get_stats()` (`index_preload_enabled`, `index_preload_time_ms`).
- The dense HNSW index continues to be warmed by the existing `_warm_vector_index()`; preload covers the keyword index only. Only worth enabling on slow network volumes.

**Bug fixed (caught via `nexus up` e2e, not unit tests):** a 0-result keyword query clobbered `last_search_timing` with a bare `{backend_ms, rerank_ms}` dict when the indexed backend returned empty — dropping `index_load_ms` for precisely the zero-result queries the issue measures (hybrid was unaffected). The keyword path now preserves the backend's merged leg timings.

## Test Plan
- [x] `pytest tests/unit/bricks/search tests/unit/server/routers tests/integration/services/test_search_router.py` → **276 passed, 13 skipped**
- [x] `ruff check` / `ruff format --check` / `mypy` clean on all 8 source files (also enforced by pre-commit)
- [x] **Live `nexus up --build` Postgres stack:**
  - `/search/stats` → `index_preload_enabled: true`, `index_preload_time_ms` non-zero (preload ran on `PgFtsBackend`)
  - `latency_breakdown.index_load_ms` present for keyword + hybrid, on empty and non-empty results
  - `request_completed` log binds `search_index_load_ms` / `search_keyword_ms` / `search_backend_ms`
  - keyword 0-result regression confirmed fixed live (`None` → populated)

## Notes
- Adds a one-line `NEXUS_SEARCH_INDEX_PRELOAD` passthrough to `nexus-stack.yml` (production Docker wiring, mirroring `NEXUS_SEARCH_PAGE_BM25`).
- Out of scope (pre-existing, separate tickets — not touched): `parsers.list` surface-coverage gap (#4135/#4187) and the `_AgentRegistryProxy.count_by_state` eviction error (#4268).

---

## Also in this PR — fix for the ~5s RPC latency this issue's e2e surfaced

Validating #4269 end-to-end (`scripts/test_build_perf_e2e.py` §8, gate `RPC latency p50 < 50ms`) failed at ~5002ms. Root-caused to a **separate gRPC-transport bug** (not the search path): the cluster VFS `read` handler overrode `timeout_ms == 0 -> 5000`. Since `timeout_ms` only governs DT_PIPE/DT_STREAM blocking (regular-file reads use `vfs_lock_timeout_ms()` and ignore it), this turned every internal pipe consumer's intended **non-blocking** poll (`timeout_ms=0` — skeleton/zoekt/task-dispatch/audit all do this) into a **5s blocking read run synchronously on a zone-mgr tokio worker**. A few consumers polling at once saturated the small worker pool shared by **raft consensus + every VFS RPC**, freezing it ~5s at a time and stalling unrelated requests.

- **Fix** (`ccfa6c2b6`): pass `timeout_ms` through unchanged, honoring the kernel contract (`0` = O_NONBLOCK, `>0` = block N ms). Regular-file reads unaffected. Adds a regression test (fails at 5.004s with the bug, passes at 0.06s without).
- **Hardening** (`b35740eab`, `ac545bf37`): route the blocking VFS handlers (`read`/`write`/`call`/`watch`/`delete`/`rename`/`copy`/`lock`/`batch_read`/`batch_write`) through `spawn_blocking` so blocking kernel work (VFS-lock waits, 30s `sys_watch`) can never starve the raft-shared async runtime.

**Verified** in the demo Docker stack: `edit(preview)` p50 **4988ms -> ~8ms**, 0 stalled across 480+ calls; transport suite 19/19; clippy clean; search unaffected. Touches only `rust/transport/src/grpc.rs` (zero overlap with the search changes above).


elfenlieds7 (changes requested):
## Kernel-Team Audit — PR #4271

### BLOCKER — `rust/transport/src/grpc.rs` is kernel-team territory

All changes to `rust/transport/src/grpc.rs` must be **removed from this PR**. This file is part of `nexusd-cluster` — the VFS gRPC service impl that runs on the raft-shared tokio runtime. Per kernel-team policy, non-kernel authors cannot modify:
- Kernel internals, kernel ABI, or kernel primitives
- nexusd-cluster behavior (services, drivers, transport, startup, packaging)

**What this PR does in grpc.rs:**
1. Removes the `timeout_ms == 0 -> 5000` override in the `read` handler
2. Adds a new `run_blocking` helper wrapping `tokio::task::spawn_blocking`
3. Wraps **12 VFS handlers** (read/write/delete/rename/copy/lock/watch/stream_read/batch_read/batch_write/call) in `spawn_blocking`
4. Adds a regression test

**Why these changes are correct but must come from kernel team:**
- The timeout fix is a valid bug fix (0 = O_NONBLOCK, the override broke pipe consumers)
- `spawn_blocking` is a sound hardening (prevent runtime starvation from blocking syscalls)
- But this is a **major change to the cluster transport execution model** — moving all VFS handlers from inline async to the blocking thread pool changes performance characteristics, concurrency model, and error propagation (`JoinError`)
- A search-timing PR should not carry kernel transport changes. Please file a separate issue and we (kernel team) will land these ourselves with proper testing and review.

---

### MEDIUM — DRY violation: `_TIMING_LEG_KEYS` duplicated

`federated_search.py` defines its own `_TIMING_LEG_KEYS` tuple:
```python
# Defined locally to avoid a federated_search → daemon import dependency.
_TIMING_LEG_KEYS = ("backend_ms", "embed_ms", "keyword_ms", ...)
```

The canonical leg keys are already in `daemon.py` (the `_SEARCH_TIMING_KEYS` tuple). Having two copies means they can drift — one gets a new leg, the other doesn't. The "import dependency" justification is wrong: `federated_search.py` already imports from the search brick package, and a constant tuple import carries zero coupling.

**Fix**: Import from `daemon.py` or extract to a shared `search_types.py`.

---

### OK — search brick changes (no kernel concerns)

- Search daemon timing (`index_load_ms`, `fallback_ms`) — well-designed, additive
- Index preload (`_preload_search_index`) — opt-in, bounded timeout, fail-soft. Good contract.
- Backend protocol `timing=` kwarg — backward-compatible (keyword-only, defaults to None)
- Router structlog binding (`_bind_search_phase_timings`) — fine, observability
- `nexus-stack.yml` env passthrough — fine
- Lifespan env wiring — fine
- Test coverage — good

---

### Summary

| Area | Verdict |
|------|---------|
| `rust/transport/src/grpc.rs` | **BLOCKER** — remove from this PR, kernel-team will land separately |
| `_TIMING_LEG_KEYS` duplication | MEDIUM — import from canonical source |
| Search timing + preload | OK |
| Router + lifespan | OK |
| Tests | OK |

elfenlieds7:
## Kernel Team Audit — PR #4271 (search preload + gRPC transport fixes)

**Reviewer:** elfenlieds7 (kernel team)
**Verdict: APPROVED**

### Kernel code changes

**Zero kernel code modified** — `git diff` on `rust/kernel/` is empty.

### Part 1: Search timing + preload (Python only) — OK, no kernel concern

All changes in `src/nexus/bricks/search/` — pure service-layer. No kernel boundary, no syscall, no HAL. Correctly layered.

### Part 2: gRPC transport bug fix + hardening — **good catch, approve**

#### 2a. `timeout_ms == 0 → 5000` override removed — **critical bug fix**

The old `if req.timeout_ms == 0 { 5000 }` override silently turned every non-blocking pipe poll (`timeout_ms=0`) into a 5s blocking read on a raft-shared tokio worker. This caused runtime starvation (~5s stalls) under concurrent pipe consumers.

Fix: pass `timeout_ms` through unchanged, honoring the kernel contract (`0` = O_NONBLOCK). Correct — regular-file reads ignore this value (use `vfs_lock_timeout_ms` from kernel/io.rs). Regression test included.

#### 2b. `run_blocking` offload — **correct hardening**

All blocking VFS handlers (`read`/`write`/`delete`/`rename`/`copy`/`lock`/`watch`/`call`/`batch_read`/`batch_write`/`stream_read_at_blocking`) now offload to `tokio::task::spawn_blocking`, keeping the raft-shared async workers free.

| Aspect | Verdict |
|--------|---------|
| Kernel code modified | **None** |
| sys_setattr/syscall signature changed | **No** |
| New kernel ABI | **No** |
| Kernel contract honored | **Yes** — `timeout_ms=0` is now correctly O_NONBLOCK |
| Transport stays thin RPC→syscall bridge | **Yes** — `run_blocking` wraps, doesn't modify kernel calls |
| No unwrap/panic in prod code | **Correct** — `spawn_blocking` JoinError mapped to `Status::internal` |

No issues found.

elfenlieds7 (approved):
## Kernel-Team Audit — PR #4271 (revised)

### Blocker from previous review — RESOLVED

The `rust/transport/src/grpc.rs` changes (timeout fix + spawn_blocking) were extracted to a separate kernel-team PR #4276 and already merged. When this PR rebases onto develop, the grpc.rs diff will vanish (develop already has it).

**Action needed before merge:** Rebase onto develop to drop the grpc.rs diff. If there are conflicts, accept develop's version (ours).

### Audit findings (Python-only changes)

**1. KERNEL-ARCHITECTURE.md consistency:** No kernel changes — search timing is entirely within the search brick and server router. No boundary violations.

**2. Forbidden zones:** No kernel ABI / syscall / hook / internal / nexusd-cluster changes. All changes are in `src/nexus/bricks/search/`, `src/nexus/server/`, and tests.

**3. DRY:** Fixed in our earlier commit (DRY — single SSOT for `_BACKEND_LEG_TIMING_KEYS`). The three-copy drift was caught and consolidated.

**4. Contract simplification:** `timing=` kwarg on `keyword_search` is keyword-only + defaults to None — backward-compatible, clean contract extension. Index preload is opt-in (env var, default off), bounded (30s timeout), fail-soft. Good.

**5. Perf:** Preload trades idle boot time for warm first query. Only on slow network volumes. Default off. Correct tradeoff.

**6. Architecture:** Observability-only change (phase timing + structlog binding). No data path changes. No new state.

### Approved

Conditional on rebase to drop grpc.rs overlap.

PR #4318, opened 2026-06-05
ci: add proto sync check for password_vault.proto drift

elfenlieds7:
## Summary

`password_vault.proto` exists in both repos (nexus `rust/services/proto/` and nexus-vfs `proto/`). nexus-vfs is SSOT. Two different teams maintain the repos — drift risk is real.

Adds `scripts/check_proto_sync.sh` that fetches the nexus-vfs SSOT and diffs against the local copy, failing on any mismatch. Can be wired into CI.


elfenlieds7:
Closing — wrong approach. The correct DRY fix is to delete the service protos from nexus-vfs (they don't belong there), keeping nexus/rust/services/proto/ as SSOT. See nexi-lab/nexus-vfs#24.

PR #4324, opened 2026-06-06
fix(docker): pin nexusd-cluster rev — edge E2E gets a stale pre-bridge-2 cluster

windoliver:
## Problem
Edge E2E smoke test fails at mount setup (run [27043206733](https://github.com/nexi-lab/nexus/actions/runs/27043206733/job/79823766626)):
```
BackendError: DT_MOUNT backend_type='remote' was acknowledged but not installed
(created=false); the gRPC server likely predates the #4262 backend-params bridge.: /
```

## Root cause
The Dockerfile installs the cluster with `cargo install --git nexus-vfs` **without `--rev`**. The cargo-install RUN layer is cached against an unchanged command string, so a **stale pre-bridge-2 `nexusd-cluster`** is reused across builds even after nexus-vfs main advances (note the 14s "Build & push" = cache reuse). The stale cluster acks DT_MOUNT but never installs it; the bridge-2-expecting server then errors at mount setup.

nexus-vfs main **does** have the bridge-2 handler (`transport/src/grpc.rs`) — it just wasn't being rebuilt.

## Fix
Pin `--rev` (`ARG NEXUSD_CLUSTER_REV`) to a current bridge-2 rev. Makes the cluster build reproducible **and** busts the stale cache on any bump.

## Validation
Built the slim cluster at the pinned rev: compiles clean, carries the bridge-2 DT_MOUNT handler (`backend_params` / `ObjectStoreProvider` strings present). A bridge-2 cluster installs DT_MOUNT correctly (verified separately against live remote + S3 mounts).

## Note
For full consistency the nexus-vfs `Cargo.lock` git pin (workspace crates / cdylib) could be bumped to the same rev in a follow-up; this PR is the minimal CI-unblocking fix. One Dockerfile line, no kernel change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,12 +80,29 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 # ---------- Install nexusd-cluster binary from nexus-vfs (Issue #3125, #4259) ----------
 # Kernel-tier Rust (including the cluster binary) migrated to the nexus-vfs
 # repo. Install the pre-built binary via `cargo install` from the git repo.
+#
+# PIN the rev. Without `--rev`, the cargo-install RUN layer is cached against an
+# unchanged command string, so a stale (pre-bridge-2 #4262) nexusd-cluster
+# binary gets reused across builds even after nexus-vfs main advances. That
+# stale cluster acks DT_MOUNT but never installs it (`created=false`), and the
+# bridge-2-expecting server fails: "DT_MOUNT backend_type='remote' was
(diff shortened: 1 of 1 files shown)


windoliver:
Folding into #4323 — both edit the same Dockerfile cluster-install line (#4324 pins --rev to fix the stale-cluster cache bug; #4323 adds --features driver-s3). Keeping them as one change avoids a guaranteed merge conflict. The --rev pin + its rationale now live in #4323.

PR #3462, opened 2026-03-28
perf: cross-zone lease-based cache invalidation (replace best-effort pub/sub)

windoliver:
## Summary

Closes #3396. Replaces fire-and-forget pub/sub cross-zone cache invalidation with a durable, lease-based protocol inspired by DFUSE (arXiv:2503.18191), closing the **5-minute stale cache window** for cross-zone permission changes.

### New components
- **`DurableInvalidationStream`** — Redis Streams (`XADD`/`XREADGROUP`/`XACK`) with pipelined drain (batch 100), bounded queue (10K), concurrent consumer (`Semaphore(10)`), dead-letter queue (`XAUTOCLAIM` + DLQ stream), and PEL health monitoring
- **`ReadFence`** — Per-zone monotonic watermark on the **read path** (~50-100ns overhead) that detects stale cached results without blocking the write path
- **`DistributedLeaseManager`** — Dragonfly-backed `LeaseManagerProtocol` with Lua atomic acquire for cross-zone lease coordination
- **`ChannelCodec`** — Shared pipe-delimited key encoder replacing fragile colon-split parsing (handles zone IDs with colons)
- **`CoordinatorConfig`** — Dataclasses grouping 15+ constructor params into 6-7 (`InvalidationChannels`, `DatabaseCallbacks`, `RecomputeCallbacks`)

### Refactors
- **CacheCoordinator**: `_dispatch_to_layer()` DRY helper consolidates 3 identical callback/stream dispatch patterns; constructor accepts `CoordinatorConfig` with backward-compat for legacy params
- **PubSubInvalidation**: migrated to `ChannelCodec`
- **ReBACPermissionCache.get()**: read fence staleness check before cache lookup

### Integration
- **Server lifecycle**: `_startup_durable_invalidation()` in `permissions.py` creates and starts durable stream + read fence, registers consumer-side handler that triggers local cache invalidation on cross-zone events, wires `DistributedLeaseManager`
- **Health endpoint**: `/health/detailed` includes PEL monitoring via `XINFO GROUPS` and read fence stats
- **Shutdown**: durable stream stopped before NexusFS close

### Architecture decisions (16 decisions reviewed interactively)
| Area | Decision |
|------|----------|
| Durable channel | Redis Streams via Dragonfly (zero new deps) |
| Revocation model | Async + read fence (write path stays fast) |
| Distributed leases | Dragonfly-backed with Lua atomic acquire |
| Phase ordering | Durable channel + read fence first (front-loads HIGH-severity fix) |
[...]

PR #4341, opened 2026-06-09
fix(#4338): self-heal rebac durable consumer on NOGROUP after cache flush

### LINKED ISSUE #4338: rebac durable consumer: NOGROUP crash loop when stream/group missing - auto-create on start
## Summary
The ReBAC durable consumer crash-loops with `NOGROUP` when its Redis/Dragonfly stream + consumer group are missing, instead of creating them.

## Evidence (prod logs, 2026-06-09, repeating)
```
redis.exceptions.ResponseError: NOGROUP No such key 'rebac:durable|root|all' or consumer
group 'zone:root' in XREADGROUP with GROUP option
```
Context: our Dragonfly instance is **volumeless** (cache tier). A Railway region move on 2026-06-02 restarted it empty — expected for a cache — and the consumer has crash-looped since.

## Asks
1. `XGROUP CREATE ... MKSTREAM` (idempotent) on consumer start; treat empty cache as normal.
2. Document what is lost when the durable stream resets, and whether a backfill/replay is needed for ReBAC correctness after a flush.

windoliver:
## Summary

Fixes the `NOGROUP` crash loop from #4338: the ReBAC durable consumer created its stream + consumer group only in `start()`. When the volumeless Dragonfly cache restarted empty mid-run (Railway region move, 2026-06-02), publishers' `XADD` re-created the stream but never the group, so every `XREADGROUP` failed with `NOGROUP`, was swallowed by the generic error handler, and retried identically forever.

- **Self-heal (ask 1):** `_consume_loop` now detects `NOGROUP`, re-runs the idempotent `XGROUP CREATE ... MKSTREAM` (`id=0`, so anything still in the stream is consumed), and resumes immediately. Re-create failures fall back to the existing 1s-backoff error path, so a half-up Redis can't hot-loop it. New `group_recreates` counter in `stats()` for observability. Boot-time creation already existed (since #3462) and is now pinned by tests.
- **Flush semantics doc (ask 2):** module docstring documents what a cache flush loses and why **no backfill/replay is needed**: relationship tuples live in the relational source of truth — the stream only carries cache-invalidation hints. Lost events mean remote zones can serve stale cached decisions only until TTL expiry (L1 result cache 300s grants / 60s denials, permission leases 30s), after which decisions re-derive from the source of truth.

## Tests

Batch 8 test pruning (5b38629f9) removed the mock-heavy cross-zone suite, leaving `DurableInvalidationStream` with no functional coverage, so the regression tests live in a new focused file `tests/unit/core/test_durable_stream_flush_recovery.py` with a purpose-built fake that mirrors real Redis error semantics (raises `NOGROUP`/`BUSYGROUP`/`ERR` like the server does).

- [x] `test_consumer_self_heals_after_cache_flush` reproduces the prod loop (verified red against unfixed source — fails in the NOGROUP loop with the prod-identical error — and green with the fix)
- [x] Pin tests: `start()` on an empty cache creates stream+group via MKSTREAM; `BUSYGROUP` on restart is swallowed
- [x] ruff check/format clean; mypy clean on changed source; pre-commit hooks pass

Closes #4338

### DIFF
--- a/src/nexus/bricks/rebac/cache/durable_stream.py
+++ b/src/nexus/bricks/rebac/cache/durable_stream.py
@@ -19,7 +19,32 @@ The sync publish preserves the synchronous invalidate_for_write() contract.
 The small window between queue-append and XADD is covered by the read fence
 (cached results are compared against the watermark, not the stream directly).
 
-Related: Issue #3396 (decisions 1A, 6A, 14A, 15A, 16A)
+Cache-flush semantics (Issue #4338):
+    The stream lives in the cache tier, often a volumeless Redis/Dragonfly
+    whose restart legitimately wipes the stream, consumer groups, PEL, and
+    DLQ.  An empty cache is treated as normal:
+
+    - start() creates the group via XGROUP CREATE ... MKSTREAM (BUSYGROUP
+      means it already exists — fine).
+    - The consumer self-heals mid-run: when XREADGROUP fails with NOGROUP,
+      the group is re-created at id="0" (consume everything still in the
+      stream) and reading resumes.  Tracked by the ``group_recreates`` stat.
+
+    What a flush loses — and why no backfill/replay is needed:
+
+    - ReBAC relationship tuples are unaffected.  The relational store is
+      the source of truth; the stream only carries cache-invalidation
+      hints.
+    - Invalidation events not yet consumed (including PEL retries and DLQ
+      entries) are lost.  Remote zones may serve stale cached decisions
+      until their caches expire — bounded by the L1 result-cache TTL
+      (default 300s grants / 60s denials) and the permission-lease TTL
+      (default 30s) — after which decisions are re-derived from the source
+      of truth.  Correctness converges without replay.
+    - Consumer-group offsets are lost, but they referenced wiped entries;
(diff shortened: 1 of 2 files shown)


windoliver:
**E2E verification** (`nexus up --build` from this branch, isolated stack, real Dragonfly):

1. **Boot on empty cache**: `[DurableStream] Started for zone root (group=zone:root)` — group created via MKSTREAM, no NOGROUP at boot.
2. **Prod incident replay** — `redis-cli FLUSHALL` on the live Dragonfly mid-run:
   ```
   03:17:16        FLUSHALL
   03:17:16.154    [DurableStream] Stream/group zone:root missing (cache flush?) — re-creating
   ```
   Exactly one warning, group re-created the same second, zero NOGROUP tracebacks (pre-fix this looped every ~1s indefinitely — the #4338 prod logs).
3. **Delivery after heal**: manual `XADD` invalidation event → consumed + ACKed in <5s (`last-delivered-id` = message id, `pending` 0, `lag` 0); handler ran clean.
4. **Repeatability**: second `FLUSHALL` → identical single-warning heal (2 warnings total for 2 flushes since boot, logs otherwise quiet).

Issue #3871, opened 2026-04-24
feat: multi-zone tokens for hub mode (follow-up to #3784)


windoliver:
## Context

Follow-up to #3784 (hub mode). The MVP that #3784 shipped binds each token to a single zone (`APIKeyModel.zone_id`). The issue examples showed `--zones eng,ops` (plural); multi-zone is deferred until the schema + auth path support it.

## Scope

- Schema change: add `zones` join table or `zone_ids` JSON column on `api_keys`.
- Update `resolve_mcp_operation_context` (`src/nexus/bricks/mcp/auth_bridge.py`) to return a zone set rather than a single zone_id.
- Update downstream search/read paths to accept a zone set filter (`WHERE zone_id IN (...)`).
- Update audit record schema (current `zone_id` field becomes `zone_ids` or "active" single).
- RLS policy review for per-zone Postgres row-level security.
- CLI: `nexus hub token create --zone eng --zone ops` (repeated flag) instead of single `--zone`.

## Out of scope

OAuth key entity + federated zone sharing — those live elsewhere.

## Acceptance

- [ ] `nexus hub token create --zone eng --zone ops --name alice` succeeds.
- [ ] MCP requests from that token see resources in both zones.
- [ ] Existing single-zone tokens keep working (migration default: wrap zone_id in a one-element set).
- [ ] Audit record preserves the zone set.

PR #4353, opened 2026-06-10
fix(#4352): surface create-key zone-guard ValueError as structured 400

### LINKED ISSUE #4352: auth-keys: POST /api/v2/auth/keys returns 500 for newly bootstrapped Koodle team zone
## Summary

Koodle production cannot use Global Files for newly provisioned teams because Nexus returns a plain `500 Internal Server Error` from `POST /api/v2/auth/keys` when creating a non-admin agent key for a newly bootstrapped zone.

This blocks Koodle's `/api/files` list and upload paths: Koodle resolves a per-team Nexus API key before constructing `NexusGlobalFileStore`, and the resolver cannot create one for the new team zone.

## Impact

- Koodle `/dashboard/files` authenticates successfully but shows `Unable to load files: Internal server error`.
- `GET /api/files?limit=50` returns 500.
- `POST /api/files` upload returns 500.
- Koodle API logs show `NexusServerError: HTTP 500` from the Nexus client.
- New Koodle team zones cannot get `nexus_team_keys`, so global files are unavailable.

## Reproduction Evidence

Target Koodle/Nexus zone:

```text
015769dc-2049-4286-9677-97e61e52edfb
```

Nexus health is green:

```json
{"status":"healthy","service":"nexus-rpc","enforcePermissions":true,"enforceZoneIsolation":true,"hasAuth":true}
```

Zone endpoints report auth provider unavailable, which Koodle already handles with its direct DB zone fallback:

```text
GET /api/zones/{zone_id} -> 503 {"detail":"Authentication provider not configured"}
POST /api/zones -> 503 {"detail":"Authentication provider not configured"}
```

The zone row exists and is active in Nexus DB:

```json
{
  "zone_id": "015769dc-2049-4286-9677-97e61e52edfb",
  "phase": "Active",
  "deleted_at": null,
  "finalizers": "[]",
  "indexing_mode": "all"
}
```

But creating a key for that zone fails:

```text
POST /api/v2/auth/keys
body: {
  "label": "Koodle team 015769dc-2049-4286-9677-97e61e52edfb",
  "user_id": "015769dc-2049-4286-9677-97e61e52edfb",
  "zone_id": "015769dc-2049-4286-9677-97e61e52edfb",
  "subject_type": "agent",
  "subject_id": "015769dc-2049-4286-9677-97e61e52edfb",
  "is_admin": false
}
-> 500 text/plain "Internal Server Error"
```

The failure is tied to the target zone, not the subject id:

```text
zone_id=015769dc-... + known working subject -> 500
zone_id=d8707207-... + subject_id=015769dc-... -> 201
```

A known older Koodle zone still allows key creation:

```text
POST /api/v2/auth/keys zone_id=d8707207-5663-4679-bbcd-ef4281d0a6c1 -> 201
```

## Koodle-side checks already done

- Required Koodle env vars are set: `NEXUS_URL`, `NEXUS_API_KEY`, `NEXUS_DATABASE_URL`, `NEXUS_TEAM_KEY_ENCRYPTION_KEY`.
- Koodle direct DB fallback successfully creates/updates the Nexus `zones` row.
- Direct SQL rollback probes for `entity_registry`, `api_keys`, and `api_key_zones` all succeed for the target zone, so the target row shape is not rejected by DB constraints.
- A direct DB-inserted API key using Nexus's documented default HMAC secret is rejected with `401`, so Koodle cannot safely bypass the REST key-creation API without Nexus's actual API-key HMAC secret.

## Expected

`POST /api/v2/auth/keys` should either:

1. successfully create a scoped non-admin key for an active zone, or
2. return a structured 4xx/5xx JSON error with the actual cause.

## Actual

It returns a plain unstructured `500 Internal Server Error`, and Koodle cannot provision per-team Nexus keys for new zones.

## Notes

This may be related to the deployment's mixed auth mode:

- `/api/v2/auth/keys` list works, so DB key storage exists.
- `/api/zones` returns `Authentication provider not configured`.
- Create works for at least one older existing zone but fails for the newly active zone above.

windoliver:
## Summary

Fixes #4352: `POST /api/v2/auth/keys` returned an unstructured plain-text `500 Internal Server Error` when the target zone was missing/not-Active from the server's DB view. Hardened through a 7-round adversarial review loop (Codex, approved round 7).

### Root cause

The #3871 zone-lifecycle guard in `DatabaseAPIKeyAuth.create_key` raises `ValueError` when the target zone is missing, Terminating/Terminated, or soft-deleted. The REST route never caught it (only `NexusError`/`NexusExchangeError` have app handlers), so Starlette's `ServerErrorMiddleware` emitted the plain-text 500 Koodle observed.

### Changes

**REST (`auth_keys.py`)**
- `handle_admin_create_key` domain `ValueError`s → `HTTPException(400, detail)`, matching sibling v2 routers.

**Validation (`database_key.py`)**
- Guards extracted to `DatabaseAPIKeyAuth.validate_key_params` (subject_type allow-list + zoneless-non-admin + zone-active), called inside `create_key`.
- Zone select takes `with_for_update` so concurrent terminate serializes with issuance (no-op SQLite, row lock Postgres).

**Side-effect safety (`admin.py` + `entity_registry.py`) — review-loop findings**
- Entity registration moved **after** the key commit: failed creates never touch the registry. Previously a rejected request committed an `entity_registry` row parented to the bad zone, and first-parent-wins kept the wrong parent forever.
- Registration is best-effort (registry is ID disambiguation; junction rows are the access SSOT per #3871), registers create_key's effective subject (`subject_id or user_id`), user/agent only.
- New `EntityRegistry.register_entity_if_absent` → `(entity, created)` with PK-conflict resolution (concurrent insert resolves to the winner, never raises).
- Behavior change: service-subject keys no longer 400 on record-store deployments (registration skipped instead of failing the request).

**Tests**
- 10 new tests: structured-400 guards, no-stale-row invariants (ValueError + non-ValueError failures), pre-existing-row survival, effective-subject registration, best-effort registration, registry provenance semantics incl. lost-insert race.
- `test_auth_keys_grants.py` fixture seeds `root` zone — file had been failing on develop since the guard landed (e74b719dd).
- Surface matrix regenerated (line drift).

### Verification

- 105 tests passing across zone-guard/registry/grants/junction/admin-handler/CLI suites.
- E2E on live `nexus up --build` stacks, before AND after (see comments): unfixed `:edge` reproduces the exact prod plain-text 500; this branch returns structured 400s, 201 for direct-SQL-inserted Active zones (Koodle's fallback path), issued keys authenticate, and failed creates leave zero registry rows.

### Accepted residuals (documented in review comment)

Zone-teardown-side lock ordering (teardown subsystem; keys fail closed at auth) · REST grants-failure-after-commit leaves a benign self-healing registry row (proper fix = grants-flow unification, follow-up).

### Note for the Koodle prod symptom

The guard firing means nexus-rpc's database does not see zone `015769dc…` as Active even though Koodle's direct-DB probes do — check that Koodle's `NEXUS_DATABASE_URL` and nexus-rpc point at the same database (ParadeDB vs rollback Postgres). After deploy, the 400 detail confirms/denies this directly.

Closes #4352

### DIFF
--- a/docs/surface-coverage/api-rpc-surface-coverage.yaml
+++ b/docs/surface-coverage/api-rpc-surface-coverage.yaml
@@ -1501,7 +1501,7 @@ operations:
   transports:
     http:
       name: GET /api/v2/auth/keys/{key_id}
-      source: src/nexus/server/api/v2/routers/auth_keys.py:351
+      source: src/nexus/server/api/v2/routers/auth_keys.py:358
   profiles:
     lite: supported
     sandbox: supported
--- a/src/nexus/bricks/auth/providers/database_key.py
+++ b/src/nexus/bricks/auth/providers/database_key.py
@@ -24,6 +24,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+VALID_SUBJECT_TYPES = ["user", "agent", "service"]
+
 
 class DatabaseAPIKeyAuth(AuthProvider):
     """Database-backed API key authentication with expiry and revocation.
@@ -266,18 +268,20 @@ class DatabaseAPIKeyAuth(AuthProvider):
         return hmac.new(secret.encode("utf-8"), key.encode("utf-8"), hashlib.sha256).hexdigest()
 
     @classmethod
-    def validate_zone_for_key(
+    def validate_key_params(
         cls,
         session: Session,
         *,
         zone_id: str | None,
         is_admin: bool,
+        subject_type: str = "user",
     ) -> None:
-        """Validate zone constraints for key issuance (#3871).
+        """Validate subject/zone constraints for key issuance.
 
-        Raises ValueError when the request is for a zoneless non-admin key
-        (round 4) or when the target zone is missing, not Active, or
-        soft-deleted (rounds 3+6). Surfaces a controlled ValueError instead
+        Raises ValueError when the subject_type is outside the allow-list,
+        when the request is for a zoneless non-admin key (#3871 round 4),
+        or when the target zone is missing, not Active, or soft-deleted
+        (#3871 rounds 3+6). Surfaces a controlled ValueError instead
         of (a) an opaque IntegrityError from the junction FK constraint or
         (b) a returned raw key that the lifecycle gate immediately rejects
         at first authentication (already persisted/displayed once).
@@ -286,6 +290,11 @@ class DatabaseAPIKeyAuth(AuthProvider):
         registration in handle_admin_create_key) must run this first so a
         rejected request leaves no side effects (#4352 review).
         """
+        if subject_type not in VALID_SUBJECT_TYPES:
+            raise ValueError(
+                f"subject_type must be one of {VALID_SUBJECT_TYPES}, got {subject_type}"
+            )
+
         # Zoneless non-admin: the token would have no zone access at auth
         # time (and downstream routes would coerce the missing zone to
         # ROOT_ZONE_ID, silently granting root).
@@ -326,13 +335,11 @@ class DatabaseAPIKeyAuth(AuthProvider):
         """
         from nexus.storage.models import APIKeyModel
 
-        final_subject_id = subject_id or user_id
-        valid_subject_types = ["user", "agent", "service"]
-        if subject_type not in valid_subject_types:
-            raise ValueError(
-                f"subject_type must be one of {valid_subject_types}, got {subject_type}"
-            )
+        cls.validate_key_params(
+            session, zone_id=zone_id, is_admin=is_admin, subject_type=subject_type
+        )
 
+        final_subject_id = subject_id or user_id
         zone_prefix = f"{zone_id[:8]}_" if zone_id else ""
         subject_prefix = final_subject_id[:12] if subject_type == "agent" else user_id[:8]
         random_suffix = secrets.token_hex(16)
@@ -340,8 +347,6 @@ class DatabaseAPIKeyAuth(AuthProvider):
 
         raw_key = f"{API_KEY_PREFIX}{zone_prefix}{subject_prefix}_{key_id_part}_{random_suffix}"
         key_hash = cls._hash_key(raw_key)
-
-        cls.validate_zone_for_key(session, zone_id=zone_id, is_admin=is_admin)
 
         api_key = APIKeyModel(
             key_hash=key_hash,
--- a/src/nexus/bricks/auth/providers/database_key.py
+++ b/src/nexus/bricks/auth/providers/database_key.py
@@ -307,7 +307,13 @@ class DatabaseAPIKeyAuth(AuthProvider):
         if zone_id:
             from nexus.storage.models import ZoneModel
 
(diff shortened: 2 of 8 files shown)


windoliver:
E2E-verified on a live `nexus up --build` stack (demo preset, database auth, image built from this branch):

| Scenario | Before (`:edge`) | After (this branch) |
|---|---|---|
| Issue repro: create key, unknown zone `015769dc…` | **500 plain-text** `Internal Server Error` (server log shows the exact `ValueError` from `database_key.py:323`) | **400 JSON** `{"detail": "...zone ... is not active (missing, Terminating, or soft-deleted)..."}` |
| Koodle flow: zone row inserted via direct SQL (their fallback), then create key | n/a | **201**, key issued |
| Issued key authenticates | n/a | `GET /files/list` 200; with editor grants write/read round-trip OK |
| Zone flipped to `Terminated` via SQL, create key | n/a | **400 JSON**, same structured detail |
| Zone restored to `Active`, create key | n/a | **201** (error message's 'create or restore' guidance is accurate) |
| List / revoke keys | n/a | 200 / 200 |

The before-run also confirms the prod diagnosis live: the unfixed image emits exactly the plain-text 500 Koodle reported, with the zone-guard `ValueError` in the traceback.

Side find while testing: ReBAC write denial on `/api/v2/files/write` returns 500 instead of 403 (builtin `PermissionError` not caught) — filed separately as #4354.

windoliver:
Adversarial review loop (Codex, 7 rounds) — **approved on round 7**. Findings fixed along the way, each with a TDD regression test:

1. **Stale entity_registry row on failed create** (high): `register_entity` committed before `create_key`'s zone guard, leaving the subject permanently parented to a bad zone (first-parent-wins). 
2. **subject_type allow-list gap** (high): `subject_type="zone"` passed registry validation but failed create_key after the row committed.
3. **Compensation ownership race** (high): own-insert tracking via pre-read could claim (and cascade-delete) a row a concurrent request inserted → added `EntityRegistry.register_entity_if_absent` with PK-conflict-resolved insert provenance.
4. **Final design — registration moved after the key commit** (round 5): failed creates now never touch the registry, eliminating the entire compensation mechanism and its races. Post-commit registration is best-effort (registry is ID disambiguation; junction rows are the access SSOT per #3871) and self-heals on the next create. Side effect: service-subject keys no longer 400 on record-store deployments.
5. **Wrong principal registered** (medium): registry derived `entity_id` from `user_id` for non-agent subjects while create_key authenticates `subject_id or user_id` — now mirrored.
6. **Zone select hardening**: `with_for_update` so concurrent terminate serializes with issuance (no-op SQLite, row lock Postgres).

**Accepted residuals** (documented, not fixed here): zone-teardown-side lock ordering (teardown subsystem; issued keys fail closed at auth); REST grants-failure-after-commit leaves the benign self-healing registry row (deleting it would reintroduce the adoption race; proper fix is unifying the REST/RPC grants flows — happy to file as follow-up if wanted).

Test coverage: 105 passing across zone-guard/registry/grants/junction/admin-handler/CLI suites; 10 new tests on this branch.

Issue #3706, opened 2026-04-10
Benchmark: readdir latency at scale + N+1 permission checks


windoliver:
## What

Two related concerns on listing:

1. `metadata.list(prefix=...)` with no limit loads all children into a Python list (`nexus_fs.py` ~4546). Unbounded memory and latency.
2. `has_accessible_descendants` called once per subdirectory in `search_service.py` ~748-752 — N separate permission checks instead of one batch.

## Benchmark

**Readdir scaling (no ReBAC):**
- List directory with 100, 500, 1K, 5K, 10K entries
- Measure wall time and peak memory
- Compare `details=True` vs `details=False` (details triggers `is_implicit_directory` per entry)

**Readdir + ReBAC:**
- Same directories with ReBAC enabled
- Measure permission check count (should be visible in Tiger cache stats)
- Compare listing a flat dir (1K files) vs a dir with 100 subdirs (triggers `has_accessible_descendants` per subdir)

## Data source

HERB corpus — `enterprise-context/` has 30 product directories each with subdirs. For 10K scale, generate: `for i in range(10000): nx.write(f"/bigdir/file_{i}.txt", b"content")`.

## Suspicious code

- `src/nexus/core/nexus_fs.py` ~4546-4547 — unbounded `metadata.list`
- `src/nexus/bricks/search/search_service.py` ~748-752 — per-dir `has_accessible_descendants`
- `src/nexus/bricks/search/search_service.py` ~456-464 — `list_directory_entries` fast path never used (method doesn't exist on any metastore)

PR #4358, opened 2026-06-11
fix(#4342): bound TigerCacheManager.initialize() boot wait

### LINKED ISSUE #4342: nexusd boot hangs indefinitely in WIRED tier — TigerCacheManager.initialize(); NEXUS_DISABLE_PERF_OPTIMIZATIONS=true bypasses
## Symptom

Fresh nexusd boot never binds its port and never logs again after:

```
2026-06-09 22:14:39,610 [INFO] nexus.bricks.search.search_service: [SearchService] Initialized
2026-06-09 22:14:39,610 [INFO] nexus.bricks.share_link.share_link_service: [ShareLinkService] Initialized
<silence — 15+ min, port 2026 never bound, [BOOT:WIRED] summary never reached>
```

Process alive (50 threads, main thread S-state), no outbound SYN_SENT, no exception (the WIRED-tier try/excepts never fire). Reproduced **2/2** boots on our prod node (Railway, image `edge@sha256:5d953bad…`, paradedb record store, ~9.4k docs, ReBAC enabled).

## Bisection

In this image's `factory/_wired.py`, construction order after ShareLinkService is WorkspaceManager → ContextBranchService → **TigerCacheManager** (`initialize()` runs synchronously at construct time). Setting `NEXUS_DISABLE_PERF_OPTIMIZATIONS=true` — which in this image gates **only** `TigerCacheManager.initialize()` (verified: the env var appears in no other module) — made the very next boot complete in ~2 min and the server has been healthy since. So the indefinite block is inside `initialize()`: `sync_resource_map()` (default-on via `NEXUS_SYNC_TIGER_RESOURCE_MAP`) or `start_worker()`.

## Context that may matter

This was the first boot of this node since 2026-06-02. Deltas accumulated in between:
- activity.db (146 GB + 50 GB WAL, 365M rows — see #4336) deleted while stopped, `NEXUS_ACTIVITY_ENABLED=false` set (the hang occurs **before** the activity-disabled log line, so not caused by it)
- search daemon has a head-of-line poisoned mutation (#4338 wedge, "refusing to checkpoint" retry loop)
- dragonfly cache was flushed (volume-less region move) — NOGROUP consumer-group errors since

Couldn't get a stack: Railway drops SYS_PTRACE (no py-spy / remote_exec). We've set `PYTHONFAULTHANDLER=1` on the service, so if it recurs we can SIGABRT and attach full thread stacks.

## Ask

`sync_resource_map()` (and anything else in `initialize()`) should be bounded/fail-soft at boot, like the search daemon's startup preload (`_PRELOAD_TIMEOUT_SECONDS` pattern in `bricks/search/daemon.py` — its header explicitly says boot must not block indefinitely). A hung perf optimization should degrade, not take the whole write/read plane down with it.

## Workaround (live on our prod)

`NEXUS_DISABLE_PERF_OPTIMIZATIONS=true`

windoliver:
Fixes #4342.

## Problem

nexusd boot hung indefinitely in the WIRED tier: `TigerCacheManager.initialize()` runs `sync_resource_map()` synchronously on the boot thread with no deadline. With `recursive=True`, `sys_readdir` bypasses the kernel fast path and hits the metastore DB, then every file pays SELECT + INSERT-ON-CONFLICT + COMMIT on a fresh pooled connection. A wedged DB call **blocks instead of raising**, so the fail-soft `try/except` never fires, the `[BOOT:WIRED]` summary is never reached, and the port is never bound (prod: 2/2 boots, 15+ min silence after ShareLinkService init).

## Fix

Mirrors the search daemon's `_PRELOAD_TIMEOUT_SECONDS` startup contract (`bricks/search/daemon.py`: boot must not block indefinitely on a perf optimization):

- Resource-map sync + optional cache warm now run on a `tiger-init-sync` **daemon thread**; boot joins it for at most `NEXUS_TIGER_INIT_TIMEOUT_SECONDS` (default 30, `0` = don't wait), then logs a warning and proceeds. Permission checks fall back to the slow path until the background sync completes — same degraded mode `NEXUS_DISABLE_PERF_OPTIMIZATIONS=true` produces permanently.
- Healthy boots keep today's semantics: the sync finishes within the window and the resource map is populated before `initialize()` returns.
- `stop_worker()` now also stops an in-progress sync between entries (`_sync_stop` event checked per readdir entry) and joins the sync thread.
- The named thread also makes the sync visible in `PYTHONFAULTHANDLER` dumps if the underlying DB wedge recurs.

## Testing

New `tests/unit/bricks/rebac/test_tiger_cache_manager.py` (5 tests), TDD red-green verified — with the src change stashed, the three new-behavior tests fail (initialize blocks on a wedged readdir, sync not stoppable); with it, 5/5 pass:

- `initialize()` returns even when `sys_readdir` blocks forever
- healthy case populates the map before returning (existing semantics)
- a sync outliving the boot timeout still completes in background
- `NEXUS_DISABLE_PERF_OPTIMIZATIONS=true` still skips entirely
- `stop_worker()` halts a grinding background sync

Broader `tests/unit/bricks/rebac/` + `tests/unit/core/test_rebac.py`: 42 passed. All pre-commit hooks (ruff, ruff-format, mypy, boundary checks) pass.

### DIFF
--- a/docs/operations/tiger-cache-boot.md
+++ b/docs/operations/tiger-cache-boot.md
@@ -57,3 +57,9 @@ when the database unwedges, the sync finishes on its own — no restart needed.
   `NEXUS_DISABLE_PERF_OPTIMIZATIONS=true`.
 - The `tiger-init-sync` thread is named, so it is visible in
   `PYTHONFAULTHANDLER=1` dumps if you need to see where it is blocked.
+- **Shutdown caveat**: graceful stop signals the sync between per-path
+  upserts, but the initial `sys_readdir` listing is materialized in one call
+  (Rust-side, see Issue #3706) and a single wedged DB statement cannot be
+  interrupted from Python. In that case shutdown logs
+  `tiger-init-sync still running ... abandoning daemon thread` and proceeds;
+  the daemon thread cannot block process exit.
--- a/docs/operations/tiger-cache-boot.md
+++ b/docs/operations/tiger-cache-boot.md
@@ -1,0 +1,59 @@
+# Tiger Cache Boot Behavior
+
+Issue #4342 bounds the Tiger Cache resource-map sync at boot. Before the fix,
+`TigerCacheManager.initialize()` ran the sync synchronously on the boot thread
+with no deadline — a wedged metastore/DB call blocked instead of raising, so
+nexusd never bound its port and never logged again after the WIRED tier began.
+
+## What runs at boot
+
+When `NEXUS_ENABLE_TIGER_CACHE=true` (and the record store is PostgreSQL),
+the WIRED tier starts a `tiger-init-sync` daemon thread that:
+
+1. Syncs `tiger_resource_map` from existing metadata (full recursive listing,
+   one upsert per path), gated by `NEXUS_SYNC_TIGER_RESOURCE_MAP`.
+2. Optionally warms the Tiger Cache (`NEXUS_WARM_TIGER_CACHE`).
+
+Boot waits for that thread at most `NEXUS_TIGER_INIT_TIMEOUT_SECONDS`, then
+proceeds. On a healthy database the sync finishes inside the window and the
+log order is unchanged:
+
+```
+Synced 155 resources to Tiger resource map
+[BOOT:WIRED] 12/13 services ready (0.904s)
+```
+
+If the sync is still running when the window closes, boot continues and logs:
+
+```
+[WARNING] Tiger resource map sync still running after 30s; continuing boot
+without it — permission checks use the slow path until the background sync
+completes (Issue #4342)
+```
+
+Permission checks fall back to the slow path (same behavior as
+`NEXUS_DISABLE_PERF_OPTIMIZATIONS=true`) until the background sync completes;
+when the database unwedges, the sync finishes on its own — no restart needed.
+
+## Environment knobs
+
+| Variable | Default | Effect |
+|----------|---------|--------|
+| `NEXUS_ENABLE_TIGER_CACHE` | `false` in the stack template, `true` in-code | Construct the Tiger Cache at all (requires PostgreSQL) |
+| `NEXUS_TIGER_INIT_TIMEOUT_SECONDS` | `30` | Max time boot waits for the resource-map sync; `0` = don't wait (pure background) |
+| `NEXUS_SYNC_TIGER_RESOURCE_MAP` | `true` | Run the resource-map sync at boot |
+| `NEXUS_WARM_TIGER_CACHE` | `false` | Also warm the cache during the sync thread |
+| `NEXUS_DISABLE_PERF_OPTIMIZATIONS` | `false` | Skip sync, warm, and worker entirely (permanent slow path) |
+| `NEXUS_ENABLE_TIGER_WORKER` | `false` | Start the legacy queue-processing worker |
+
+## Troubleshooting
+
+- **Boot warns "sync still running"** on every start: the metastore listing or
+  the per-path upserts are slow or blocked. Check PostgreSQL for lock waiters
+  on `tiger_resource_map` and for dead client connections. The server is fully
+  functional meanwhile; only permission-check latency is degraded.
+- **Symptom of the original bug** (pre-fix images): silence after the
+  ShareLinkService init line, port never bound. Workaround on old images:
+  `NEXUS_DISABLE_PERF_OPTIMIZATIONS=true`.
+- The `tiger-init-sync` thread is named, so it is visible in
+  `PYTHONFAULTHANDLER=1` dumps if you need to see where it is blocked.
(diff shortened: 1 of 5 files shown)


windoliver:
## E2E validation (`nexus up --build`, worktree image, shared preset + `NEXUS_ENABLE_TIGER_CACHE=true`, 151 seeded files)

**Healthy path** — sync runs inline, semantics unchanged:
```
05:24:33,810 Tiger resource map sync complete: 155 resources
05:24:33,810 Synced 155 resources to Tiger resource map
05:24:33,955 [BOOT:WIRED] 12/13 services ready (0.904s)
```

**Wedged-DB path** — reproduced the prod hang for real: emptied `tiger_resource_map`, held `LOCK TABLE ... IN EXCLUSIVE MODE` in an open psql transaction (SELECTs pass, the sync's first INSERT blocks — same blocking shape as #4342), restarted nexusd with `NEXUS_TIGER_INIT_TIMEOUT_SECONDS=5`:
```
05:30:44,216 [WARNING] Tiger resource map sync still running after 5s; continuing boot
             without it — permission checks use the slow path until the background
             sync completes (Issue #4342)
05:30:44,456 [BOOT:WIRED] 12/13 services ready (5.248s)
...lock released...
05:34:31,921 Uvicorn running on http://0.0.0.0:2026
05:34:32,673 Tiger resource map sync complete: 155 resources   <- background recovery, no restart
```
On the unfixed image this exact setup blocks `_wired()` indefinitely (the prod symptom: port never bound, 15+ min silence). With the fix: bounded at the knob value, boot proceeds, and the perf optimization self-restores once the DB unwedges — `tiger_resource_map` ended at 164 rows.

**Functional after recovery**: REST write/read OK; permission-filtered recursive list of 151 files: 163ms cold / ~21ms warm.

Second commit exposes `NEXUS_TIGER_INIT_TIMEOUT_SECONDS` as a compose passthrough (empty default → in-code 30s).

Notes from the run (pre-existing, out of scope):
- The WIRED tier runs twice per boot (two app instances), so two `tiger-init-sync` threads each did the bounded wait — harmless, both warned and proceeded.
- A *fully* write-locked `tiger_resource_map` can still stall later lifespan steps via write-through grant INSERTs (and `ACCESS EXCLUSIVE` even stalls migration introspection); this fix bounds the boot-time sync specifically, per the issue ask.

windoliver:
## Gap closure round

- **Shutdown wiring** (new commit): `TigerCacheManager.stop_worker` is now registered in `nx._close_callbacks` (same pattern as the write-observer/delivery workers), so the `tiger-init-sync` thread and queue worker stop at `NexusFS.close()` instead of being abandoned. `stop_worker` no-op/idempotent contract pinned by a new test (6 tests total in the file).
- **Ops doc** (new commit): `docs/operations/tiger-cache-boot.md` — bounded-init contract, all tiger env knobs, degraded-mode troubleshooting.
- **Correction to my earlier comment**: the "WIRED tier runs twice per boot" note was wrong — a clean restart shows exactly one `[BOOT:WIRED]` and one resource-map sync per boot. The doubled lines came from a `docker logs --since` window overlapping two boots during the wedge experiment. No duplicate-sync gap exists.
- **Follow-ups filed**: #4359 (boot-phase write-through grant INSERTs are still unbounded past the WIRED tier — observed stalling lifespan ~3 min under the table lock until release), #4360 (`sync_resource_map` N+1 per-path round trips; batch the upserts now that it's off the critical path).

windoliver:
## Adversarial review loop (Codex) — PASSED on round 5

Three hardening commits came out of rounds 1-3:

- **Round 1** (`63f7b0e0b`): single-flight `initialize()` under a lifecycle lock with a fresh per-run stop event (a manual refresh could orphan a running sync and clear a stop signal a concurrent shutdown just set); explicit warning when shutdown abandons a wedged daemon sync; materialized-list fidelity test (production `sys_readdir` returns a list — a wedged listing blocks inside the call). The reviewer's paginated-cancellable-listing recommendation was rejected as infeasible Python-side: recursive pagination re-materializes the full listing per page and the Rust engine materializes internally (#3706).
- **Round 2** (`5c77704b3`): a stop that ends the sync early no longer falls through into the DB-bound warm phase; `start_worker()` serialized under the lifecycle RLock with a closure-captured stop event (a stopped-but-slow worker could be revived by event replacement); both threads signaled before either join; `NEXUS_TIGER_INIT_TIMEOUT_SECONDS` validated finite ≥ 0 (`inf` raised OverflowError inside `Thread.join`, aborting initialize and silently skipping the worker) and clamped to `threading.TIMEOUT_MAX`.
- **Round 3** (`6ea0cf77e`): `stop_worker()` is terminal — a boot thread resuming from its bounded join after shutdown can no longer restart the worker, and post-stop `initialize()` is a logged no-op; the close callback is registered **before** `initialize()` so a `close()` racing the bounded boot wait owns the just-started sync thread.
- **Round 4**: finding ("timeout must force a Tiger read bypass") rejected with code evidence — the Tiger check path is positive-only (`manager.py` honors only `tiger_result is True`, everything else falls through to the slow path) and the sync writes only int-ID mappings, never bitmap grants, so a partial map is a cache miss, not stale authorization state. Round 5 independently confirmed: "no concrete read path where a partial resource map flips auth or listing behavior versus an empty map" → **approve**.

12 unit tests in `test_tiger_cache_manager.py` (each new behavior red-green verified), rebac suites 46/46, CI 30/30 on the latest sha.

Issue #52, opened 2025-10-16
[v0.1.0] Implement GCS backend as thin wrapper over GCS SDK


joezhoujinjing:
## Summary

Implement GCS backend as a thin wrapper over Google Cloud Storage SDK, allowing Nexus to read/write files directly to user's existing GCS buckets.

## Architecture

```
User's GCS Bucket (unchanged):
  my-bucket/
    ├── data/file1.txt        ← existing files
    ├── data/file2.txt
    └── models/model.pkl

Nexus:
  metadata.db (local):        ← tracks what we've seen
    file_paths:
      /data/file1.txt → gs://my-bucket/data/file1.txt
      /data/file2.txt → gs://my-bucket/data/file2.txt
  
  GCSBackend:                 ← thin wrapper over GCS SDK
    - read/write via bucket.blob()
    - list via bucket.list_blobs()
```

## Key Points

- ✅ **No CAS** - files stored at their actual paths in GCS
- ✅ **No metadata in GCS** - all metadata in local SQLite
- ✅ **No directory markers** - GCS uses prefixes naturally
- ✅ **User's bucket unchanged** - works with existing GCS structure
- ✅ **Simple wrapper** - just GCS SDK calls

## Implementation

```python
class GCSBackend(Backend):
    """Thin wrapper over GCS SDK."""
    
    def __init__(
        self,
        bucket_name: str,
        project_id: str | None = None,
        credentials_path: str | None = None,
        prefix: str = "",
    ):
        from google.cloud import storage
        
        if credentials_path:
            client = storage.Client.from_service_account_json(
                credentials_path, project=project_id
            )
        else:
            client = storage.Client(project=project_id)
        
        self.bucket = client.bucket(bucket_name)
        self.prefix = prefix.rstrip("/") + "/" if prefix else ""
    
    def write_content(self, path: str, content: bytes) -> str:
        """Upload content to GCS."""
        blob = self.bucket.blob(f"{self.prefix}{path}")
        blob.upload_from_string(content)
        
        # Return hash for metadata.db
        return hashlib.sha256(content).hexdigest()
    
    def read_content(self, path: str) -> bytes:
        """Download content from GCS."""
        blob = self.bucket.blob(f"{self.prefix}{path}")
        if not blob.exists():
            raise FileNotFoundError(f"Not found: {path}")
        
        return blob.download_as_bytes()
    
    def delete_content(self, path: str) -> None:
        """Delete blob from GCS."""
[...]

Issue #54, opened 2025-10-16
[v0.2.0] Implement permission operations (chmod, chown, chgrp)


windoliver:
## Overview
Implement UNIX-style permission operations: `chmod`, `chown`, and `chgrp`.

## Background
Builds on #53 (permissions data model). Once we have owner/group/mode fields, we need operations to modify them.

### API Design

Add new methods to `Embedded` class (`src/nexus/core/embedded.py`):

```python
def chmod(self, path: str, mode: int | str) -> None:
    """
    Change file mode (permissions).
    
    Args:
        path: Virtual path
        mode: Permission mode (int: 0o644, str: "rw-r--r--")
    
    Raises:
        NexusFileNotFoundError: If file doesn't exist
        PermissionError: If user doesn't have permission to chmod
        InvalidPathError: If path is invalid
    """

def chown(self, path: str, owner: str | None = None, group: str | None = None) -> None:
    """
    Change file owner and/or group.
    
    Args:
        path: Virtual path
        owner: New owner (None = no change)
        group: New group (None = no change)
    
    Raises:
        NexusFileNotFoundError: If file doesn't exist
        PermissionError: If user doesn't have permission to chown
        InvalidPathError: If path is invalid
    """

def stat(self, path: str) -> FileMetadata:
    """
    Get file metadata including permissions.
    
    Args:
        path: Virtual path
    
    Returns:
        FileMetadata with owner, group, mode, etc.
    """
```

### Permission Checking

**Who can chmod?**
- Owner can chmod their own files
- Admin can chmod any file
- Others cannot chmod

**Who can chown?**
- Only admin can chown (security: prevent privilege escalation)
- Owner cannot give away their files

### Implementation Tasks

- [ ] Add `chmod()` method to `Embedded`
  - [ ] Support int mode (0o644)
  - [ ] Support string mode ("rw-r--r--", "644")
  - [ ] Validate mode value (0-0o777)
  - [ ] Check user has permission to chmod
  - [ ] Update metadata store
- [ ] Add `chown()` method to `Embedded`
  - [ ] Support changing owner only
  - [ ] Support changing group only
  - [ ] Support changing both
  - [ ] Require admin privileges
  - [ ] Update metadata store
- [ ] Add `stat()` method to `Embedded`
  - [ ] Return full `FileMetadata`
  - [ ] Include owner, group, mode
- [ ] Add `chgrp()` convenience method (calls `chown(group=...)`)

### Mode String Parsing

Support multiple formats:
[...]

PR #4368, opened 2026-06-11
chore(deps): bump nexus-vfs to d78e4ae — durable kernel metastore (#4343)

### LINKED ISSUE #4343: nexus-cluster: LocalMetaStore lives in tempfile::tempdir() — entire VFS namespace lost on every restart (set_metastore_path never wired)
## Summary

`nexus-cluster` (the Rust kernel all profiles spawn via gRPC since the KernelClient migration) keeps the **entire VFS namespace** (`LocalMetaStore`, redb) in a **`tempfile::tempdir()`** that is never swapped for a durable path. Every kernel restart silently drops every file registration; payload bytes remain on disk but `exists` → `false` and reads → 404 for everything written before the boot.

This was the actual root cause behind #4339 (closed — the historical-data salvage it asked for is done downstream). Filing this separately so the live defect has its own tracker: **any nexus deployment using the cluster kernel loses its whole namespace on every restart.**

## Mechanism (source)

- `rust/kernel/src/kernel/mod.rs` (~690–700, nexus-vfs): bare kernels boot with a tempfile-backed `LocalMetaStore` — "The tempdir is held by the kernel; it drops when the kernel drops (or when `set_metastore_path` swaps in a real path)."
- `Kernel::set_metastore_path` (the only durable wiring) is **never called in production**: in `nexus-vfs` the only callers are `benches/syscall_bench.rs` and `benches/readdir_bench.rs`; `rust/profiles/cluster/src/main.rs` never wires it.
- Python side: `src/nexus/remote/kernel_client.py` stubs `set_metastore_path` as a no-op with the comment "For subprocess mode, this was already passed via env" — but no env var carries a metastore path to the binary; `NEXUS_DATA_DIR` only sets the payload/object-store dir.

## Production confirmation (twice, same night)

- A path returning `exists=true` before a restart returns `exists=false` after it; a canary written after the boot returns `true`. Visible namespace == "writes since last boot", exactly.
- Files sit byte-identical in the live `<data_dir>/root/<path>` tree while the API 404s them.
- Downstream evidence + recovery log: DeepBuildAI/koodle#1302.

## Repro (one line)

write → restart `nexus-cluster` → `GET /api/v2/files/exists` for the written path must stay `true` (today it returns `false`).

## Asks

1. Wire a durable metastore in the cluster profile at boot — e.g. `set_metastore_path("<NEXUS_DATA_DIR>/metastore.redb")` (or honor an explicit `NEXUS_METASTORE_PATH`), before serving traffic.
2. Add the restart-survival regression test above to CI.
3. (Nice to have) a `walk-import` admin command that re-registers everything found under `<data_dir>/root/**` — it's the universal recovery path for instances already bitten (we run exactly this downstream via `batch/write`, 15,817 files in ~5 min).

Happy to send a PR for (1)+(2) if maintainers prefer.

windoliver:
Carries [nexus-vfs#42](https://github.com/nexi-lab/nexus-vfs/pull/42) (merged as `d78e4ae`) into the product: the cluster kernel finally persists its VFS namespace across restarts — fixes #4343, the failure mode that wiped all file visibility on koodle prod twice on 2026-06-09.

**Changes**
- `Cargo.toml`: all 7 `nexus-vfs` workspace rev pins `32c4731` → `d78e4ae` (+ `Cargo.lock` regenerated — picks up the ed25519 chain from plugin-signing #52, also in range).
- `Dockerfile`: `ARG NEXUS_VFS_REV` → `d78e4ae` so image builds compile the fixed `nexusd-full`.

**Validation**
- `cargo check --workspace` green at the new rev (235 crates).
- The kernel change itself was E2E-validated downstream against a docker stack built at this rev's predecessor head: write → container restart → `exists` stays true; `ephemeral` opt-out reproduces the old amnesia; empty env refuses to boot (evidence: DeepBuildAI/koodle#1302).

**Operator note**: the metastore override env is `NEXUS_KERNEL_METASTORE_PATH` — deliberately not `NEXUS_METASTORE_PATH`, which the Python server sets for its own legacy metadata path and copies into the kernel subprocess (details in the arg docs / nexus-vfs#42 discussion).

Close #4343 once this merges; the remaining nice-to-have from that issue (an upstream walk-import admin command for already-bitten instances) is split out separately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,7 +168,7 @@ dependencies = [
 [[package]]
 name = "backends"
 version = "0.1.0"
-source = "git+https://github.com/nexi-lab/nexus-vfs?rev=32c47310e48d449bb5c3f02e48ef12455bf0e8fa#32c47310e48d449bb5c3f02e48ef12455bf0e8fa"
+source = "git+https://github.com/nexi-lab/nexus-vfs?rev=d78e4ae83de68b6d024f1561674622650089640c#d78e4ae83de68b6d024f1561674622650089640c"
 dependencies = [
  "ahash",
  "base64",
@@ -214,6 +214,12 @@ name = "base64"
 version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
+
+[[package]]
+name = "base64ct"
+version = "1.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2af50177e190e07a26ab74f8b1efbfe2ef87da2116221318cb1c2e82baf7de06"
 
 [[package]]
 name = "bincode"
@@ -363,6 +369,12 @@ checksum = "ea0095f6103c2a8b44acd6fd15960c801dafebf02e21940360833e0673f48ba7"
 
 [[package]]
 name = "const-oid"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
+
+[[package]]
+name = "const-oid"
 version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a6ef517f0926dd24a1582492c791b6a4818a4d94e789a334894aa15b0d12f55c"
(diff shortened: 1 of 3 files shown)


windoliver:
CI failures root-caused: #42's fail-closed mount rule rejected the password vault's bootstrap shape (`/vault` as the very first mount on a bare kernel, empty router) — all 19 `nexus-vault` failures trace to `no parent route for non-root mount /vault`. Fix: [nexus-vfs#54](https://github.com/nexi-lab/nexus-vfs/pull/54) allows the empty-router bootstrap (validated downstream: vault suite 0/19 → 19/19 via `[patch]` override against this repo). Once #54 merges I'll re-bump this PR's pins to the new rev and CI should go green.

Issue #87, opened 2025-10-18
v0.3.0: Skills System - Enterprise Features


windoliver:
## Overview

Implement enterprise features for Skills System: analytics, semantic search, governance, and audit trails.

### Usage Analytics
- [ ] Track skill usage count
- [ ] Calculate success rate
- [ ] Measure execution time
- [ ] Track users
- [ ] Count forks
- [ ] `SkillManager.get_skill_analytics()`

### Semantic Skill Search
- [ ] Search by description (not just name)
- [ ] Use Nexus semantic search
- [ ] Search across all skill tiers
- [ ] Limit and pagination
- [ ] `SkillManager.search_skills()`

### Skill Governance
- [ ] Approval workflow for org-wide skills
- [ ] Review process
- [ ] Approval tracking
- [ ] Only approved skills in `/shared/`

### Audit Trails
- [ ] Log all skill usage
- [ ] Track execution details
- [ ] Store findings/results
- [ ] Query audit logs
- [ ] Compliance reporting

### Analytics Dashboard
- [ ] Most used skills
- [ ] Top contributors
- [ ] Success rates by skill
- [ ] Average execution times
- [ ] Organization-wide metrics

## Implementation

**Files to create**:
- `src/nexus/skills/analytics.py` - Analytics tracking
- `src/nexus/skills/governance.py` - Approval workflows
- `src/nexus/skills/audit.py` - Audit logging

**Database tables**:
```sql
CREATE TABLE skill_usage (
    id UUID PRIMARY KEY,
    skill_name TEXT,
    agent_id TEXT,
    execution_time FLOAT,
    success BOOLEAN,
    timestamp TIMESTAMP
);

CREATE TABLE skill_approvals (
    id UUID PRIMARY KEY,
    skill_name TEXT,
    submitted_by TEXT,
    status TEXT,
    reviewers TEXT[],
    approved_at TIMESTAMP
);
```

## Testing

- [ ] Test usage tracking
- [ ] Test semantic search
- [ ] Test approval workflow
- [ ] Test audit logging
- [ ] Test analytics queries

## Success Criteria

- ✅ Analytics tracked for all skill usage
- ✅ Semantic search finds relevant skills
- ✅ Approval workflow works
- ✅ Audit logs complete and queryable
- ✅ Dashboard shows accurate metrics

Issue #82, opened 2025-10-18
v0.2.0: FUSE Mount Performance Optimizations


windoliver:
## Overview

Implement performance optimizations for FUSE mount including caching, read-ahead, and lazy loading to ensure smooth user experience.

### Attribute Caching
- [ ] Implement TTL cache for `getattr()` results
- [ ] Cache size: 1024 entries, TTL: 60s
- [ ] Invalidate on write/delete operations
- [ ] Cache hit/miss metrics

### Content Caching
- [ ] LRU cache for file content
- [ ] Cache recently read files
- [ ] Size limit (e.g., 100 files or 100MB)
- [ ] Eviction policy
- [ ] Cache parsed content separately

### Read-Ahead Prefetching
- [ ] Predict next file to be read
- [ ] Prefetch directory siblings
- [ ] Sequential read patterns
- [ ] Async prefetch in background
- [ ] Configurable prefetch strategy

### Lazy Loading
- [ ] Don't load file content until accessed
- [ ] Lazy parse binary files
- [ ] Parse on first read, not on write
- [ ] Background parsing queue
- [ ] Prioritize actively accessed files

### Batch Operations
- [ ] Batch metadata lookups
- [ ] Use `batch_get_content_ids()` for directory listings
- [ ] Reduce N+1 queries
- [ ] Batch cache invalidation

### Memory Management
- [ ] Configurable cache sizes
- [ ] Memory limits
- [ ] Graceful degradation under memory pressure
- [ ] Monitor memory usage

## Implementation

**Files to update**:
- `src/nexus/fuse/mount.py` - Add caching layer
- `src/nexus/fuse/cache.py` - Cache implementations

**Cache Libraries**:
- `cachetools` - TTL and LRU caches
- Or implement custom caches

## Configuration

```yaml
# nexus.yaml
fuse:
  attr_cache_size: 1024
  attr_cache_ttl: 60
  content_cache_size: 100
  content_cache_max_mb: 100
  enable_readahead: true
  enable_lazy_loading: true
```

## Testing

- [ ] Benchmark without caching
- [ ] Benchmark with caching
- [ ] Measure cache hit rates
- [ ] Test memory usage under load
- [ ] Test read-ahead effectiveness
- [ ] Test with large files
- [ ] Test with many files
- [ ] Profile bottlenecks

## Performance Targets

| Operation | Target | Current | Improvement |
|-----------|--------|---------|-------------|
| `getattr()` | <1ms | TBD | TTL cache |
[...]

PR #4446, opened 2026-06-27
chore(deps): E2E-gate nexus-vfs PR #87 (reconstruct_global_path root-cause fix) at branch HEAD

elfenlieds7:
## Summary

E2E-gating PR for **nexus-vfs PR #87** — the actual root-cause fix for the same-zone DT_MOUNT placeholder wrong-path bug that surfaced (but was misattributed) in reverted PR #82 (sys_mkdir/rename/setattr defer-to-peer) and closed Phase 3a (sys_mkdir side-effect).

Pinned at #87's BRANCH HEAD (\`b32598b17\`), NOT main yet. Same gating pattern Phases 1 & 2 used.

## What I found (via local Docker Desktop repro)

\`reconstruct_global_path\` (rust/raft/src/distributed_coordinator.rs:1837) used \`.min()\` over \`global_path\` to pick from \`cross_zone_mounts[parent_zone_id]\`. Lexicographic min is an arbitrary heuristic. It happens to work when there's exactly ONE entry (CI's typical 2-node topology with a single \`/shared → sharedzone\` gateway). When there are multiple, it picks an arbitrary one — and when that's NOT the root gateway, the same-zone placeholder gets installed at a wrong path.

Reproduced directly: joiner's wire_mount_core installs at \`/shared/cc-tasks/joiner/cc-tasks/founder\` instead of \`/shared/cc-tasks/founder\`. Joiner FUSE then sees nothing at \`/mnt/cc-tasks/founder/\`. Every \`host_task_write\`-seeded cross-node test fails.

## Why PR #82 + Phase 3a got the blame

Both PRs added kernel call sites that perturbed apply order enough to fan out \`cross_zone_mounts[sharedzone]\` to multiple entries. The bug was always there; the kernel changes just made it observable.

## Architectural self-check

- **Simple contract** — pure-function fix, no syscall/plugin/HAL ABI change.
- **No boundary leak** — fix at federation tier where the data is.
- **SSOT** — root-gateway entry IS the SSOT; was being queried via heuristic.
- **DRY** — single fix at single site.
- **Rust-first.**
- **Perf** — O(n) over tiny vec; no allocation change.
- **Raft contract** — no raft surface change; \`cross_zone_mounts\` is a node-local derived cache.
- **Systemic** — addresses gateway-selection semantics at the function that owns the contract. Future kernel-side perturbations no longer regress E2E for this reason.

## Flow

1. ✅ nexus-vfs #87 opened — kernel-tier CI in flight
2. **THIS PR** — pinned at \`b32598b17\`
3. Wait for Federation E2E + cc-tasks-share E2E
4. **If green** → admin-merge #87, re-pin THIS PR at main SHA, admin-merge this; queue Phase 3a re-attempt
5. **If red** → diagnose locally; fix in place; force-push #87; re-bump pin

## Test plan

- [ ] \`Federation E2E (Docker)\` green
- [ ] \`cc-tasks-share E2E (Docker)\` green
- [ ] All other CI gates green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -218,7 +218,7 @@ dependencies = [
 [[package]]
 name = "backends"
 version = "0.1.0"
-source = "git+https://github.com/nexi-lab/nexus-vfs?rev=fc16c58661583cbad934a7cb87716734ee951409#fc16c58661583cbad934a7cb87716734ee951409"
+source = "git+https://github.com/nexi-lab/nexus-vfs?rev=b32598b1716568e0e33c7a4ad8c59661595eb323#b32598b1716568e0e33c7a4ad8c59661595eb323"
 dependencies = [
  "ahash",
  "base64",
@@ -550,7 +550,7 @@ checksum = "3d52eff69cd5e647efe296129160853a42795992097e8af39800e1060caeea9b"
 [[package]]
 name = "contracts"
 version = "0.1.0"
-source = "git+https://github.com/nexi-lab/nexus-vfs?rev=fc16c58661583cbad934a7cb87716734ee951409#fc16c58661583cbad934a7cb87716734ee951409"
+source = "git+https://github.com/nexi-lab/nexus-vfs?rev=b32598b1716568e0e33c7a4ad8c59661595eb323#b32598b1716568e0e33c7a4ad8c59661595eb323"
 dependencies = [
  "parking_lot",
  "serde",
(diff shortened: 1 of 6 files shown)


elfenlieds7:
Closing — nexus-vfs PR #87 root-cause hypothesis was wrong (see close comment). Develop stays at Phase 2 baseline; no kernel perturbation while the deeper bug is understood.

PR #4553, opened 2026-07-31
feat(search): query-conditional post-fusion recency decay (#4543)

### LINKED ISSUE #4543: search: recency has zero influence on ranking — add post-fusion time decay (query-conditional)
## Problem

Recency has zero influence on ranking. Nothing in the query path reads any timestamp — a 3-year-old chunk and yesterday's chunk with equal text relevance tie. Deployments indexing living corpora (notes, mail, docs) systematically surface stale material.

The data is already free to carry: `file_paths.updated_at` is in the covering-index INCLUDE list of the join every backend already performs (`src/nexus/storage/models/file_path.py:96`).

(Refs pinned at `bd52314f7`.)

## Proposal

1. Select `fp.updated_at` in the three backend SELECTs (pg_fts, pg_vector, sqlite) and carry `mtime` on `BaseSearchResult`.
2. Post-fusion multiplicative boost in `_search_via_backends` after `daemon.py:2128-2131`:
   `score *= 1 + w * H / (H + age_days)` (hyperbolic decay; `H` = half-life days).
   Default **off** (`w=0`). Request params `recency_weight` / `recency_half_life_days`, plus `DaemonConfig` defaults.
3. Optional `recency=auto`: the QueryRouter already computes temporal intent (`TEMPORAL_WORDS`, `src/nexus/bricks/search/query_router.py`) — enable the boost only for time-framed queries.

Hard rule: never put decay in SQL `ORDER BY` — it corrupts the rank positions RRF consumes, and the vector arm's HNSW scan must stay pure-distance to remain index-usable.

## Acceptance

- Fixture with near-duplicate content at different mtimes: temporal query ranks newer first with recency on; default-off request byte-identical.
- Per-result attribution field (e.g. `recency_boost`) set only when the boost fired.

## Dependencies

Soft: reuse the per-request knob plumbing from the fusion-params issue (land that first). Hard: none.

windoliver:
Closes #4543.

Post-fusion multiplicative recency boost `score *= 1 + w·H/(H+age_days)` at a
single chokepoint in `SearchDaemon.search()`, hydrating `file_paths.updated_at`
with one index-only batch query (the existing covering index already carries
it, routed through the daemon owner loop for asyncpg loop affinity). Default
off; `recency=on|auto` per request or `NEXUS_SEARCH_RECENCY` per deployment —
`auto` fires only on queries containing a new `RECENCY_WORDS` intent word
(deliberately not `TEMPORAL_WORDS`, whose "history"/"before" often want OLD
docs). Per-result `recency_boost` attribution, omit-when-None so default-off
responses are byte-identical. Federated: knobs included in the cache key,
forwarded to local zones only (rrf_k RPC precedent). No SQL ORDER BY changes;
RRF rank inputs and the HNSW scan stay pure. Fail-soft: hydration errors log,
increment `recency_attach_failures` (exposed in `get_stats()`), and return
unboosted results.

Known limitations (documented in the spec): `updated_at` bumps on any row
update including re-index bookkeeping; local-vs-remote boost asymmetry in
federated raw-score merge; graph branch ignores the knobs (same pre-existing
gap as alpha/fusion); `"new"` in RECENCY_WORDS can fire on lexical false
positives in auto mode.

Design: docs/superpowers/specs/2026-07-31-search-recency-decay-design.md
Plan: docs/superpowers/plans/2026-07-31-search-recency-decay.md

Tests: 136 passed across recency unit suites, daemon fusion-params/page-pooling
regression pins, router param tests, federated integration, protocol
conformance, and the surface freshness gate.


windoliver:
Live-stack e2e (built from this branch via `nexus up --build`, demo preset):

- **Recency feature e2e: 14/14.** Near-duplicate docs with a 1500-day-backdated `updated_at`: default request byte-identical (no `recency_boost` keys, no hydration query); `recency=on&recency_weight=1.0` ranks the fresh doc first with boost 2.0 (fresh) / 1.02 (stale) — exact formula values; `auto` fires on "latest …" and stays off for neutral queries; bogus mode → 400, out-of-bounds weight → 422; `/search/stats` exposes `recency_attach_failures: 0`.
- **Zone-scoped path (the owner-loop asyncpg concern): verified.** Search in the `research` zone with `recency=on` boosts all results and the fail-soft counter stays 0 — hydration runs on the owner loop as intended, not degrading silently.
- **`scripts/test_build_perf_e2e.py`: 26 passed, 0 failed** (HERB hit rate 8/8, RPC p50 9ms), so the feature introduces no regression across CRUD/search/permissions/auto-index/delete-propagation.

This also closes the review's deferred item "real-DB hydration SELECT untested" — real Postgres, real covering index, real path shapes.

windoliver:
**Default changed to `auto`** (commits 4af4781ab + 019d56051): zero-config deployments now boost recency-intent queries out of the box — nothing to set at deploy time. Neutral queries stay byte-identical via the intent gate; opt out with `NEXUS_SEARCH_RECENCY=off` or per-request `recency=off`. The compose template exposes the three `NEXUS_SEARCH_RECENCY*` knobs.

Re-verified live on a fresh `nexus up --build` (container has NO recency env — pure in-code default):

- **Recency e2e: 18/18**, now including: zero-config "latest …" query boosts and reorders (fresh doc above a 1500-day-stale near-duplicate); zero-config neutral query byte-identical; explicit `recency=off` opts out even for intent queries.
- **`scripts/test_build_perf_e2e.py`: 26 passed, 0 failed** under default-auto (one CLI-transient SIGPIPE flake on an intermediate run, clean on rerun) — HERB retrieval and permission-filtered search unaffected.
- Local sweep: 485 passed across search unit + integration suites; spec updated (decision summary, config table, acceptance note documenting the default-off → default-auto revision).

One test-side finding worth knowing (documented limitation, not a bug): the indexer's bookkeeping writes bump `file_paths.updated_at`, so an mtime fixture backdated before indexing settles gets re-freshened — the e2e now waits for indexing to quiesce before backdating.

Issue #18, opened 2025-10-12
[v0.2.0] Implement PDF parser with text and table extraction


windoliver:
## Description
Implement PDF parser that extracts text, images, tables, and metadata from PDF files.

## Architecture Reference
See NEXUS_COMPREHENSIVE_ARCHITECTURE.md Section: "Rich Format Parsing" - PDF Parser

### Text Extraction
- Extract all text from PDF
- Preserve formatting where possible
- Handle multi-column layouts
- Support scanned PDFs (OCR) - optional for v0.2.0

### Table Extraction
- Detect tables in PDF
- Extract as structured data (rows/columns)
- Support merged cells
- Export as CSV or JSON

### Image Extraction
- Extract embedded images
- Save with original format
- Extract image metadata

### Metadata Extraction
- Title, author, subject
- Creation/modification dates
- Page count
- PDF version

### PDF Parser Class
- [ ] Create `PDFParser` class implementing `Parser` interface
- [ ] Use `pypdf` for basic text extraction
- [ ] Use `pdfplumber` for table extraction
- [ ] Support multi-page PDFs

### Text Extraction
- [ ] Extract text page by page
- [ ] Handle different encodings
- [ ] Preserve paragraph structure
- [ ] Remove headers/footers (optional)

### Table Extraction
- [ ] Detect tables using `pdfplumber`
- [ ] Extract table data
- [ ] Return as list of dicts
- [ ] Handle malformed tables

### Image Extraction
- [ ] Extract images using `pypdf`
- [ ] Save as separate files or embed in result
- [ ] Support JPEG, PNG formats
- [ ] Store image metadata

### Metadata Extraction
- [ ] Extract PDF metadata
- [ ] Include in ParseResult
- [ ] Handle missing metadata gracefully

### Testing
- [ ] Test with various PDF types
- [ ] Test with scanned PDFs
- [ ] Test with tables
- [ ] Test with images
- [ ] Test with encrypted PDFs (should fail gracefully)

## Sample Test Files
Create `tests/fixtures/` with:
- Simple text PDF
- PDF with tables
- PDF with images
- Multi-page PDF
- Encrypted PDF

## Acceptance Criteria
- ✅ Can extract text from PDF files
- ✅ Can extract tables as structured data
- ✅ Can extract images
- ✅ Metadata extracted correctly
- ✅ Handles errors gracefully
- ✅ Tests pass with >80% coverage

## Related Files
- `src/nexus/parsers/pdf.py`
[...]

PR #4567, opened 2026-08-01
fix(search): bounded file_paths projection wait + 409 skip surfacing for POST /search/index (#4566)

### LINKED ISSUE #4566: search: POST /search/index silently drops documents whose file_paths projection row hasn't landed yet (count:0) — write-then-index always misses
## Symptom

`POST /api/v2/search/index` immediately after `POST /api/v2/files/write` for the same path returns `{"status":"indexed","count":0}` and the submitted text NEVER becomes searchable. The same request re-issued ~90s later returns `count:1` and the text is searchable within seconds.

Live repro against a production deployment of `edge@sha256:1c7aabe1…` (466a60b37), paradedb-backed, PgFts+vector:

1. `files.write /perf-test/e2e-dbg-<ts>/probe.md` content **X** ("alphaxx<ts> …") → 200
2. `search.index [{id, text: **Y** ("zuluxx<ts> …"), path: same}]` → `{"count":0,"zone_id":"root"}`
3. Poll keyword search 60s: **X (blob bytes, via mutation consumer): 1 hit. Y (explicit text): 0 hits, forever.**
4. Re-issue the exact same `search.index` at +90s → `count:1`; Y searchable ~6s later.

## Mechanism

`SearchDaemon._index_documents_on_current_loop` (daemon.py ~3690) resolves `path_id` from the **postgres `file_paths` projection**, which is populated asynchronously by the operation-log consumer. A write-then-index call always lands inside that projection lag (~5–10s observed), hits the `path_id is None` branch, and **silently `continue`s** — the doc is dropped from the count and the text is lost (no retry queue, no error).

This contradicts both docstrings on the path:
- route `search_index_documents`: *"Fails closed with HTTP 500 … so clients can retry instead of silently losing data"*
- `_index_documents_on_current_loop`: *"Failures bubble up so the HTTP boundary returns 500 (Decision #18) instead of silently returning count=0"*

…yet the no-row branch is exactly a silent count=0.

## Impact

- **Call-by-call indexing (Decision #18) is broken for freshly-written files** — the dominant client pattern (write blob → index text).
- Clients that index **extracted text for binary blobs** (PDF/DOCX/XLSX parses, OCR, sidecar markdown) lose search coverage permanently: the consumer indexes the binary bytes, the extracted text never lands, and `count:0` gives no HTTP-level failure to retry on.
- Downstream freshness regressed ~2.6s → ~10s for write-then-index clients (visibility now rides the consumer chain instead of the explicit call).

The pre-#4553-era daemon (observed on `edge@529f43e2…` / 2573d10a1) indexed the submitted text without requiring a pre-existing `file_paths` row, so this is a behavior regression for existing API consumers, not just a doc gap.

## Fix options (any of these would do)

1. **Bounded wait/retry on the missing row** inside `index_documents` (the lag is seconds; e.g. 3×2s) before giving up.
2. **Surface the skip**: return `{count, skipped: [paths]}` and/or a 409/207 so callers can retry deterministically — minimum viable fix, keeps the reader-only ownership rule ("daemon must not create file_paths").
3. Let the explicit call **stage text keyed by virtual_path** so the consumer attaches it when the row lands (preserves the provided text over blob bytes).

As a client-side stopgap we now verify `count` and retry/requeue on shortfall, but the silent-drop default will keep biting other API consumers.

/cc the #4544/#3699 lineage — the path_id resolution requirement seems to have arrived with the ChunkStore ownership refactor.

windoliver:
Closes #4566

## Problem

`POST /api/v2/search/index` immediately after `files.write` for the same path returns `{"status":"indexed","count":0}` and the submitted text never becomes searchable. `_index_documents_on_current_loop` resolves `path_id` from the `file_paths` projection, which the operation-log consumer populates asynchronously — a write-then-index call always lands inside that lag (~5–10s observed), hits the `path_id is None` branch, and silently `continue`s. This contradicts the fail-closed docstrings on both the route and the daemon method (Decision #18).

## Fix (options 1 + 2 from the issue, combined)

**Daemon — bounded projection wait.** Docs whose row is missing are collected and re-polled up to `NEXUS_SEARCH_INDEX_PATH_WAIT_ATTEMPTS` (default 5) rounds, sleeping `NEXUS_SEARCH_INDEX_PATH_WAIT_SECONDS` (default 2.0s) between rounds. Defaults cover the observed 5–10s lag. The budget is per **batch**, not per doc — worst-case added latency is `attempts × delay` regardless of batch size. A daemon with no DB session skips the wait entirely (the lookup can never succeed). Set attempts to `0` to disable the wait.

**Daemon — no more silent drops.** `index_documents` now returns `ExplicitIndexResult {indexed, skipped}`. Anything unresolved after the wait lands in `skipped` (zone-stripped virtual paths) plus a WARNING log and two new `DaemonStats` counters (`explicit_index_path_waits`, `explicit_index_skipped_no_path`). The reader-only ownership rule is preserved — the daemon never creates `file_paths` rows.

**Route — fail closed.** Any skipped doc ⇒ HTTP 409 with `detail = {error, count, skipped, zone_id}` so clients can retry deterministically. Indexing is idempotent (`replace_document_chunks` upsert), so retrying the whole batch after a 409 is safe; docs indexed before the 409 stay indexed. The 200 response shape is unchanged.

**Callers.** `mount_service` logs `indexed`/`skipped` from the result; connector wiring and the RPC `semantic_search_index` handler discard the result as before, so their best-effort semantics are unchanged (skips log instead of raising). Side benefit: skill README indexing in `connectors.py` (nx.write → immediate index) was racing the same projection lag and silently dropping — the bounded wait now covers it.

## Notes for review

- **Not addressed here:** option 3 from the issue (staging text keyed by `virtual_path` for the consumer to attach). The bounded wait narrows the race but a consumer content-index landing *after* a successful explicit index can still overwrite the explicit text with blob bytes; that ordering question is the same one the pre-#4553 daemon had and is left as a follow-up if it bites in practice.
- Invalid docs (missing/empty `text` or `path`) are still silently ignored, as before — that's malformed input, not a projection race; tightening it to a 400 would change the lenient contract mount/connector callers rely on.
- `tests/integration/search/test_connector_search_indexing.py` fails 6/6 **identically on develop** (stale pre-#3699 `_backend.upsert` expectations, same as the self-skipping `test_zone_isolation.py`); not a regression from this change.

## Testing

- New: `tests/unit/bricks/search/test_daemon_explicit_index_path_wait.py` (7 tests: row lands mid-wait, never lands → skipped, no-session fast path, attempts=0, mixed batch, stats counters, env-knob defaults/overrides).
- New: `tests/unit/server/api/v2/test_search_index_skipped_409.py` (4 tests: 200 all-indexed, 409 with skipped detail, int-returning double compat, 500 on daemon error).
- `tests/unit/bricks/search` + `tests/unit/server/api/v2` + `tests/unit/server/routers` + RPC handler test: **623 passed, 8 skipped** (pre-existing PG-URL skips).
- mypy clean on the three changed source files; surface-coverage YAML regenerated (line-ref shifts only) and validator exits 0.
- `nexus-stack.yml` passes through the two new env knobs.

### DIFF
--- a/docs/surface-coverage/api-rpc-surface-coverage.yaml
+++ b/docs/surface-coverage/api-rpc-surface-coverage.yaml
@@ -443,7 +443,7 @@ operations:
   transports:
     grpc_expose:
       name: add_mount
-      source: src/nexus/bricks/mount/mount_service.py:873
+      source: src/nexus/bricks/mount/mount_service.py:874
   profiles:
     lite: supported
     sandbox: supported
@@ -2542,7 +2542,7 @@ operations:
   transports:
     grpc_expose:
       name: delete_connector
-      source: src/nexus/bricks/mount/mount_service.py:945
+      source: src/nexus/bricks/mount/mount_service.py:946
   profiles:
     lite: supported
     sandbox: supported
@@ -2578,7 +2578,7 @@ operations:
   transports:
     grpc_expose:
       name: delete_saved_mount
-      source: src/nexus/bricks/mount/mount_service.py:1403
+      source: src/nexus/bricks/mount/mount_service.py:1404
   profiles:
     lite: supported
     sandbox: supported
@@ -4392,7 +4392,7 @@ operations:
   transports:
     grpc_expose:
       name: get_mount
-      source: src/nexus/bricks/mount/mount_service.py:1203
+      source: src/nexus/bricks/mount/mount_service.py:1204
   profiles:
     lite: supported
     sandbox: supported
@@ -4771,7 +4771,7 @@ operations:
   transports:
     grpc_expose:
       name: has_mount
-      source: src/nexus/bricks/mount/mount_service.py:1220
+      source: src/nexus/bricks/mount/mount_service.py:1221
   profiles:
     lite: supported
     sandbox: supported
@@ -5186,7 +5186,7 @@ operations:
   transports:
     grpc_expose:
       name: list_connectors
-      source: src/nexus/bricks/mount/mount_service.py:1003
+      source: src/nexus/bricks/mount/mount_service.py:1004
   profiles:
     lite: supported
     sandbox: supported
@@ -5276,7 +5276,7 @@ operations:
   transports:
     grpc_expose:
       name: list_mounts
-      source: src/nexus/bricks/mount/mount_service.py:1187
+      source: src/nexus/bricks/mount/mount_service.py:1188
     sdk:
       name: MCPClient.list_mounts
       source: src/nexus/remote/domain/mcp.py:43
@@ -5348,7 +5348,7 @@ operations:
   transports:
     grpc_expose:
       name: list_saved_mounts
-      source: src/nexus/bricks/mount/mount_service.py:1307
+      source: src/nexus/bricks/mount/mount_service.py:1308
   profiles:
     lite: supported
     sandbox: supported
@@ -5435,7 +5435,7 @@ operations:
   transports:
     grpc_expose:
       name: load_mount
-      source: src/nexus/bricks/mount/mount_service.py:1355
+      source: src/nexus/bricks/mount/mount_service.py:1356
   profiles:
     lite: supported
     sandbox: supported
@@ -7008,7 +7008,7 @@ operations:
   transports:
     grpc_expose:
       name: reauth_mount
-      source: src/nexus/bricks/mount/mount_service.py:1099
+      source: src/nexus/bricks/mount/mount_service.py:1100
   profiles:
     lite: supported
     sandbox: supported
@@ -7317,7 +7317,7 @@ operations:
   transports:
     grpc_expose:
       name: remove_mount
-      source: src/nexus/bricks/mount/mount_service.py:917
+      source: src/nexus/bricks/mount/mount_service.py:918
   profiles:
     lite: supported
     sandbox: supported
@@ -7589,7 +7589,7 @@ operations:
   transports:
     grpc_expose:
       name: save_mount
-      source: src/nexus/bricks/mount/mount_service.py:1236
+      source: src/nexus/bricks/mount/mount_service.py:1237
   profiles:
     lite: supported
     sandbox: supported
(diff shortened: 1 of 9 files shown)


windoliver:
Pushed 0e3a94be3 closing both caveats from the PR description:

**Clobber race** — anchored on the existing `indexed_content_id` tracking column instead of new staging machinery: the explicit call stamps it (CAS, only when chunks actually landed), and both mutation consumers now skip an UPSERT whose current `content_id` already matches the marker. Explicit text survives the consumer's blob indexing; a real content change advances `content_id` and indexes normally. Remaining window is only a consumer pass already in-flight when the stamp lands (noted in code); the legacy `/search/refresh` path is deliberately un-gated as an operator escape hatch.

**Stale tests** — `test_connector_search_indexing.py` rewritten for the current MountService signature + pipeline daemon (was 6/6 red on develop). 651 tests green across the affected suites; surface YAML unchanged by this commit.

windoliver:
Live-validated end-to-end with `nexus up --build` (demo preset, image built from this branch) — pushed 65ac52b3a with three defects the live run exposed:

1. **`content_id` is identity-valued on the local backend** (it's the path, unchanged across rewrites) — the equality-only covered gate would have latched forever and silently stopped re-indexing edits. Covered now also requires `last_indexed_at >= updated_at`; the stamp CAS guards on the resolution-time `updated_at`.
2. **The legacy refresh loop was the actual live clobberer** (fires on every VFS write via the post-write hook, 5s debounce — observed overwriting the explicit text exactly on schedule). Now gated.
3. **Same-tick consumer race** (fts consumer gate-checked 3ms before the stamp landed; interleaved replaces left a mixed chunk set). Closed with a bounded post-index verify-repair guard: re-replaces a raced foreign write, stands down if the row moved to a newer version.

Verification on the live stack:
- Feature e2e 6/6: write-then-index → `200 count:1` in 2.2s, text searchable in seconds, **survives the consumer chain**, blob bytes cleanly superseded, `409 {skipped}` for pathless docs.
- `scripts/test_build_perf_e2e.py`: **26 passed, 0 failed** — including section 9 (auto-index on edit) and section 10 (delete propagation), the flows most exposed to the new gate.
- 652 unit/integration tests green locally; mypy + ruff clean.

PR #4615, opened 2026-08-10
fix(search): batch_search proxy — typed per-query failures + full param forwarding (#4612)

### LINKED ISSUE #4612: search: gRPC proxy batch_search drops per-query typed failures — failed queries serialize as genuine empty results
Found while evaluating the post-pivot generation for a production upgrade (follow-on from #4610/#4611).

## Problem

The P12 pivot's Python proxy (`src/nexus/bricks/search/daemon.py`) discards the Rust plugin's per-query error field:

```python
resp = await self._get_stub().BatchQuery(req)
return [[_result_to_base(hit) for hit in sub.results] for sub in resp.responses]   # sub.error dropped
```

Meanwhile the `/api/v2/search/query/batch` router still does `isinstance(inner, BatchQueryFailure)` to emit the per-entry additive `error` field — a check the proxy can never satisfy, so it is now dead code. Every failed inner query (embedder unavailable, per-query timeout, backend error) serializes as `{"results": [], "total": 0}` with **no** `error` key.

This regresses the 2026-08-04 contract (`ce3805f38` + hardening follow-ups): *"absence of `error` means the result set is genuine"*. Downstream consumers rely on it — Koodle's sub-path batch fold (DeepBuildAI/koodle#2050/#2164) uses the per-entry error to fail coverage accounting CLOSED; with the field silently gone, a failed workspace query reads as "no matches" and cross-workspace coverage is silently overstated. This is an upgrade blocker for adopting the pivot generation in that deployment.

## Also stale

- The router docstring still advertises "All unique query texts are embedded in ONE `embed_batch` call and the inner searches run concurrently (`NEXUS_SEARCH_BATCH_CONCURRENCY`, default 8)" — pre-#4611 the plugin ran the batch serially with no shared embed; post-#4611 concurrency defaults to 4 (env `NEXUS_SEARCH_BATCH_CONCURRENCY`, clamp 1..=16) and duplicate texts pre-warm a query-embedding cache rather than one literal shared `embed_batch` call. Either bump the plugin default to 8 to match the documented contract or fix the docstring.

## Proposed fix

Map `sub.error` (when non-empty) to `BatchQueryFailure(error=...)` in the proxy's `batch_search` so the router's existing typed-failure path lights up again, restoring the additive per-entry `error` field on the wire. A round-trip test asserting a failed inner query (e.g. semantic with no embedder configured) yields `error` in the HTTP response would pin the contract.

windoliver:
Closes #4612.

## What broke

The P12 proxy's `batch_search` shipped with three defects the mock-daemon tests couldn't see (they `AsyncMock` the daemon, so the route↔proxy call contract was never exercised):

1. **Signature drift — live `/query/batch` 500s.** The route called `batch_search(fetch_specs, zone_id=...)` (dict specs + kwarg, the deleted Python daemon's shape) while the P12 proxy takes a bare positional list of typed requests and reads attributes off each. Every live call raised `TypeError` before reaching the plugin.
2. **Dropped tuning knobs.** The batch pb mapping only forwarded `q`/`zone`/`limit`/`path`/`type` — `alpha`, `fusion_method`, `rrf_k`, `expand`, and the recency params were silently discarded on the batch path while the single path forwards them all.
3. **Dropped typed failures (the issue as filed).** `sub.error` from the plugin was discarded, so a failed inner query (embedder unavailable, per-query timeout, backend error) serialized as `{"results": [], "total": 0}` with no `error` key — regressing the 2026-08-04 contract (*absence of `error` means the result set is genuine*) that fail-closed consumers (Koodle's cross-workspace batch fold) rely on for coverage accounting. The route's `isinstance(inner, BatchQueryFailure)` branch was dead code.

## Fix

- Route builds `SearchRequest` objects (zone_id inside each request — exactly the single `/query` shape) and calls `batch_search(requests)`.
- Proxy maps requests through a shared `_request_to_pb` helper used by both `search` and `batch_search`, so batch can never again silently drop a knob the single path forwards.
- Proxy maps a non-empty `sub.error` to `BatchQueryFailure`, lighting the route's existing typed-failure path back up — the additive per-entry `error` field is on the wire again.
- Stale route docstring fixed: it advertised the pre-#4611 story (one `embed_batch` call, concurrency default 8); it now describes the shipped behaviour (duplicate-text embed-cache pre-warm, `NEXUS_SEARCH_BATCH_CONCURRENCY` default 4, clamp 1..=16). Kept the plugin default at 4 — that was a deliberate #4611 choice (each in-flight query can spawn two blocking tasks).

## Tests

- Mock-daemon forwarding assertions updated to the typed shape.
- New `TestBatchRealProxyRoundtrip` class: route → **real** `SearchDaemon` proxy → fake gRPC stub → route response. Pins (a) the route↔proxy call contract (would have caught the 500), (b) per-entry `error` serialization for a failed inner query, (c) the full knob set reaching the wire. `tests/integration/services/test_search_query_batch.py`: 58 passed.
- Related suites green: `tests/unit/bricks/search`, `tests/unit/server/api/v2`, search-startup failsoft. Surface-coverage validation passes.

### DIFF
--- a/docs/surface-coverage/api-rpc-surface-coverage.yaml
+++ b/docs/surface-coverage/api-rpc-surface-coverage.yaml
@@ -7692,7 +7692,7 @@ operations:
   transports:
     http:
       name: POST /api/v2/search/expand
-      source: src/nexus/server/api/v2/routers/search.py:1611
+      source: src/nexus/server/api/v2/routers/search.py:1615
   profiles:
     lite: supported
     sandbox: supported
@@ -7716,7 +7716,7 @@ operations:
       source: src/nexus/bricks/search/search_service.py:1792
     http:
       name: POST /api/v2/search/glob
-      source: src/nexus/server/api/v2/routers/search.py:1469
+      source: src/nexus/server/api/v2/routers/search.py:1473
     mcp:
       name: nexus_glob
       source: src/nexus/config/tool_profiles.yaml:1
@@ -7743,7 +7743,7 @@ operations:
       source: src/nexus/bricks/search/search_service.py:2102
     http:
       name: POST /api/v2/search/grep
-      source: src/nexus/server/api/v2/routers/search.py:1346
+      source: src/nexus/server/api/v2/routers/search.py:1350
     mcp:
       name: nexus_grep
       source: src/nexus/config/tool_profiles.yaml:1
@@ -7796,7 +7796,7 @@ operations:
   transports:
     http:
       name: POST /api/v2/search/index
-      source: src/nexus/server/api/v2/routers/search.py:1531
+      source: src/nexus/server/api/v2/routers/search.py:1535
   profiles:
     lite: supported
     sandbox: supported
@@ -7848,7 +7848,7 @@ operations:
   transports:
     http:
       name: POST /api/v2/search/refresh
-      source: src/nexus/server/api/v2/routers/search.py:1590
+      source: src/nexus/server/api/v2/routers/search.py:1594
   profiles:
     lite: supported
     sandbox: supported
--- a/src/nexus/bricks/search/daemon.py
+++ b/src/nexus/bricks/search/daemon.py
@@ -16,6 +16,7 @@ from typing import TYPE_CHECKING, Any
 import grpc
 
 from nexus.bricks.search.results import BaseSearchResult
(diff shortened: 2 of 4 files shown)


windoliver:
Superseded by #4622 (all three adoption blockers combined into one PR per review preference).

Issue #4613, opened 2026-08-10
release: post-P12, no published image provides working search — publish the plugin-host artifact + document the deployment contract


windoliver:
Follow-on from the #4598 (P12) pivot, found while evaluating the edge image for a downstream production upgrade (see also #4612).

## Problem

Since #4598 deleted the Python SearchDaemon, the Rust `nexus-search-plugin` cdylib (hosted by `nexusd-cluster`, gRPC loopback) is the sole search backend. But the release pipeline hasn't caught up:

- `docker-publish.yml` pushes **only the Python-server image** to GHCR (`edge`/`stable`/`latest`).
- The plugin-host image is built in-workflow (`nexusd-cluster:latest`, `Dockerfile.nexusd-cluster`) **only for the e2e smoke job** — it is never published.
- The server's boot probe fail-softs when the plugin is unreachable: it logs a warning and disables search entirely (`server/lifespan/search.py`).

Net effect: **every published image tag currently ships with no working search out of the box.** A downstream deployment that tracks `edge` (single-container server topology — e.g. the Koodle production instance, pinned by digest to the last pre-P12 edge) cannot adopt any post-P12 image without an artifact that doesn't exist yet.

## Ask

1. Publish the plugin-host image (or a bundled single-container variant that supervises server + `nexusd-cluster --plugin-dir` together) to GHCR alongside the server image, from the same commit, so a deployable pairing is pinnable by digest.
2. Document the deployment contract for server deployments: required processes, `NEXUS_SEARCH_PLUGIN_TARGET` wiring, plugin signing expectations, model provisioning (`NEXUS_SEARCH_MODEL_DIR`, ORT dylib), and the migration story for corpora indexed under the previous stack (embedding model changes 1536-dim API → 384-dim local mE5 ⇒ full ANN reindex).
3. Consider gating `stable`/`latest` tags on the e2e-edge job's sidecar topology actually being reproducible from published artifacts — today the smoke passes using an image the public can't pull.

Happy to help test against our downstream deployment once an artifact exists. Blockers we're tracking for adoption: this issue + #4612 (batch typed-failure contract).

windoliver:
[...]

Issue #4610, opened 2026-08-09
search: /query/batch executes queries serially and re-embeds identical query text (post-#3699 regression)


windoliver:
## Symptom

Koodle's cross-workspace ⌘K fan-out (DeepBuildAI/koodle#2176) measures only ~30% workspace coverage inside its 7s budget against prod Nexus, with coverage COLLAPSING when client-side concurrency is raised (16 → 19%, 24 → 2%). The fan-out is throughput-bound at Nexus.

## Root cause (two parts, both in `bricks/search/daemon.py`)

**1. `_batch_search_on_current_loop` runs its queries strictly serially.** The post-#3699 fallback replaced the dedicated backend batch call with a plain `for q in queries: await self.search(...)` loop — its own comment notes the amortisation loss: *"we fan out per-query through the new backend stack instead. This is functionally equivalent for callers, just without the single-call embedding amortisation."* A caller batching N queries in one `POST /api/v2/search/query/batch` pays N × full query latency.

**2. Query embeddings are never cached.** `EmbeddingClient` is deliberately cache-free (its docstring defers caching to the daemon: *"Caching, if needed in the future, should be wired explicitly in the daemon"*) — but nothing in the daemon wires it. `_embed_query` calls `litellm.aembedding` every time. Koodle's fan-out sends the SAME query text across ~3 sub-path queries × ~60 workspaces per keystroke → ~180 identical embedding API calls per keystroke, each a full provider round-trip inside the serial loop.

## Proposed fix

- Bounded in-process LRU for query embeddings at the `_embed_query` chokepoint (keyed by model+dim+text, failures never cached, size env-tunable, 0 disables).
- `_batch_search_on_current_loop`: pre-warm duplicate query texts once, then execute the batch with bounded concurrency (env-tunable, 1 restores serial). Order-preserving; per-query failure isolation unchanged.

PR follows.

PR #4619, opened 2026-08-10
feat(search-plugin): remote/API embedding backend (#4614)

### LINKED ISSUE #4614: search-plugin: support remote/API embedding backends (post-P12 the only embedder is local mE5-small ONNX)
Third adoption blocker from the downstream (Koodle) production evaluation, alongside #4612 and #4613.

## Problem

Pre-P12, deployments embedded via `EmbeddingClient` (litellm → any API model, e.g. 1536-dim OpenAI). P12 deleted that path; the Rust plugin's only real embedder is `FastEmbedder` — local ONNX, `build_default_embedder` hardcodes the `mE5-small-v1` tag, dim probed from the model file.

For a corpus indexed and quality-tuned on API embeddings this is a hard migration wall AND a likely quality regression:

- Our production tuning (weighted fusion, alpha=0.3) was measured on API embeddings: hybrid MRR 0.620 → 0.746 on a 58-query known-item eval. mE5-small (384-dim) is a substantially weaker model on this kind of finance corpus; adopting P12 as-is means re-running that entire eval with a realistic chance the semantic lane gets worse.
- The embedding swap forces a full ANN reindex regardless — fine once, but there is currently no way to reindex onto embeddings of equal quality.

## Ask

An `Embedder` impl (the trait is already clean: `embed_batch`/`dim`/`tag`) backed by a remote embedding API — configurable model + dims + endpoint/key via env, tag derived from model+dims so the AnnIndex directory contract (`ann-<tag>-v<n>`) keeps working. Blocking HTTP inside the existing `spawn_blocking` call sites is fine; a per-process request cache already exists post-#4611.

This keeps the P12 architecture intact (plugin owns indices + embeds) while letting deployments keep their embedding quality and choose local-vs-API per environment. Happy to contribute the impl if the direction is acceptable.

windoliver:
Closes #4614.

## Why

P12 deleted the `EmbeddingClient` path (litellm → any API model); the plugin's only real embedder became local ONNX mE5-small (`build_default_embedder` hardcodes the tag). For a corpus indexed and quality-tuned on API embeddings that's a hard migration wall AND a likely quality regression — the forced reindex has nowhere equal-quality to land.

## What

`RemoteEmbedder` — an `Embedder` impl backed by an OpenAI-compatible `/v1/embeddings` endpoint, keeping the P12 architecture intact (plugin owns indices + embeds):

| Env var | Meaning |
|---|---|
| `NEXUS_SEARCH_EMBED_API_URL` | endpoint URL; presence selects the remote backend |
| `NEXUS_SEARCH_EMBED_MODEL` | model name (required with URL) |
| `NEXUS_SEARCH_EMBED_DIM` | embedding dim (required with URL — pins the ANN index without a boot-time network probe; every response validated against it) |
| `NEXUS_SEARCH_EMBED_API_KEY` | optional bearer token |
| `NEXUS_SEARCH_EMBED_TIMEOUT_SECONDS` | optional, default 30 |

Design points:

- **Fail-loud config**: URL set but model/dim missing errors in `build_default_embedder` instead of silently falling back to local ONNX — a typo must not quietly reindex the corpus into a different vector space.
- **Tag contract**: `api-<sanitized-model>-<dim>` derives from model+dim so the `ann-<tag>-v<n>` directory contract keeps working across embedder swaps (same model+dim ⇒ same vector space ⇒ same directory).
- **Blocking HTTP is safe here**: every `embed_batch` call site already runs inside `spawn_blocking`, and `reqwest::blocking` drives I/O on its own dedicated runtime thread — the plugin's tokio pool is never re-entered. reqwest 0.13 + rustls (no OpenSSL linkage; same line the tree already resolves — Cargo.lock grows 2 lines).
- **Always compiled** (not behind `semantic` — no ONNX involved): a non-semantic build gets a dense lane when an endpoint is configured. A configured remote wins over `FastEmbedder`.
- Response items assembled by per-item `index` (not arrival order); provider error bodies surfaced truncated in typed `Runtime` errors; >512-input batches transparently chunked.
- The per-process query-embedding cache (#4610/#4611) applies unchanged, so the batch fan-out pattern embeds each unique text once per process.

## Tests

Env-shape parsing via an injectable lookup (no process-env races across test threads), tag sanitization for HF-style names, and live localhost HTTP round-trips (plain `std::net::TcpListener`, no new dev-deps): ordered-by-index assembly + bearer auth on the wire, provider error status surfacing, dim-mismatch rejection, empty-batch short-circuit. Full plugin lib suite: 129 passed. `cargo fmt` clean; clippy clean on changed code (the `--all-targets` failures in `tests/*_e2e.rs` `mock_mut` predate this branch).

Deployment wiring/docs for these vars lands with #4613's deployment-contract doc (the plugin reads `nexusd-cluster` process env directly, same as `NEXUS_SEARCH_MODEL_DIR`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2637,6 +2637,7 @@ dependencies = [
  "prost",
  "protoc-bin-vendored",
  "regex",
+ "reqwest",
  "serde",
  "serde_json",
  "tantivy",
@@ -3638,6 +3639,7 @@ checksum = "219c5811de6525e5416c7d5d53bb656d3afdbc6c5af816e0802bcfa42dbdc1c3"
 dependencies = [
  "base64 0.22.1",
  "bytes",
+ "futures-channel",
  "futures-core",
  "futures-util",
  "http",
--- a/rust/services/search-plugin/Cargo.toml
+++ b/rust/services/search-plugin/Cargo.toml
@@ -110,6 +110,23 @@ ort = { version = "2.0.0-rc.10", default-features = false, features = [
 ], optional = true }
 
 
+# ── remote/API embedder (#4614) ───────────────────────────────────
+# Blocking HTTP for `RemoteEmbedder` — embeds via an OpenAI-compatible
+# `/v1/embeddings` endpoint when NEXUS_SEARCH_EMBED_API_URL is set.
+# Blocking is deliberate: every embed_batch call site already runs on
+# a spawn_blocking worker, and reqwest::blocking keeps its own
+# dedicated runtime thread so the plugin's tokio pool is never
+# re-entered.  rustls so the cdylib carries no OpenSSL linkage to
+# manage across platforms; same 0.13 line the wider tree already
+# resolves.  NOT feature-gated behind `semantic` — remote embedding
+# needs no ONNX, so a non-semantic build still gets a dense lane when
+# an API endpoint is configured.
+reqwest = { version = "0.13", default-features = false, features = [
+    "blocking",
+    "json",
+    "rustls",
+] }
+
 # Parking-lot mutex around the per-zone index writer.  tantivy's own
 # IndexWriter is Send + Sync but only allows one writer per index at a
 # time; the lock is short-held (writer.add_document + writer.commit)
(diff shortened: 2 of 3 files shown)


windoliver:
Superseded by #4622.

PR #4621, opened 2026-08-10
release: publish nexusd-cluster plugin-host image + deployment contract docs (#4613)

windoliver:
Closes #4613.

## Problem

Post-P12 (#4598) the Rust `nexus-search-plugin` cdylib hosted by `nexusd-cluster` is the **sole** search backend, but `docker-publish.yml` only published the Python-server image. The plugin-host image existed solely as an in-workflow build for the e2e smoke — so **every published tag shipped with no working search out of the box** (the boot probe fail-softs and disables search), and the smoke passed using an artifact the public can't pull.

### 1. Publish the plugin-host image (issue ask 1)

New self-contained `dockerfiles/Dockerfile.nexusd-cluster-plugins`:
- cluster binary built from the `Cargo.toml`-pinned nexus-vfs rev (`nexus-vfs-src` build context, same convention as the existing files),
- plugin cdylib built from this repo and Ed25519-signed in-build (BuildKit secret, never a layer),
- one runtime image with the signed pair in `/plugins` and `NEXUS_PLUGIN_DIR` preset.

Self-contained (vs the CI-only `Dockerfile.search-plugin-test` which layers on a locally pre-built base) so one `docker build` reproduces the published artifact with any buildx driver — that's what makes push-by-digest multi-arch publishing possible. Reciprocal KEEP-IN-SYNC notes added to the two siblings.

New `build-cluster-platform` (amd64 + arm64, native runners, push-by-digest — same pattern as the server image jobs) and `push-cluster-manifest` jobs publish `ghcr.io/nexi-lab/nexusd-cluster` with the **same tag rules** as the server image (`edge`, `stable`/`latest`/semver on tags, commit SHA) — every server tag gets a matching plugin-host tag from the same commit, pinnable by digest.

### 2. e2e-edge validates the published artifact (issue ask 3)

The sidecar steps drop the in-workflow `nexusd-cluster` + `search-plugin-test` builds (and the nexus-vfs checkout) and `docker run` the just-pushed image with the same single-node flags the compose bootstrap used (`--bind-addr 0.0.0.0:2126 --no-tls --insecure-no-auth --plugin-dir /plugins`, shared `/workspace` volume). The smoke now exercises the exact bytes the public pulls. (Full gating — pushing `stable`/`latest` only *after* the smoke — would need the tag push restructured to a post-e2e job; left as a follow-up, noted in the issue.)

### 3. Deployment contract docs (issue ask 2)

[...]

Issue #4556, opened 2026-07-31
federated: remote-zone search RPC is rejected as unknown method — registry-remote zones always fail


windoliver:
## Problem

`FederatedSearchDispatcher._search_remote_zone` sends a `search` Call RPC to registry-remote zones (`src/nexus/bricks/search/federated_search.py`), but the receiving servicer rejects it before dispatch:

- `parse_method_params` (`src/nexus/server/protocol.py:53`) raises `Unknown method: search` — `search` has no entry in `METHOD_PARAMS` (generated or overrides).
- The fallback only accepts `@rpc_expose`'d methods (`state.exposed_methods`); `search` is only in the static dispatch table (`src/nexus/server/rpc/dispatch.py:99`), which is consulted *after* param parsing.

Net effect: **every registry-remote zone lands in `zones_failed`** — remote federation over this path has never returned results. Additionally, even if dispatch succeeded, `handle_search` (`src/nexus/server/rpc/handlers/filesystem.py:117`) reads only `path`/`limit`/`search_type`, routes to `nexus_fs.search` (not the SearchDaemon hybrid pipeline), and returns a `{"results": ...}` envelope. No fusion knob (`alpha`/`fusion_method`/`rrf_k`, #4541) reaches remote zones.

Existing tests mask this by mocking `transport.call_rpc` to return a bare list.

## Proposal

- Add a param schema for a daemon-backed remote search method (or extend `SemanticSearchParams` with the fusion knobs) and route it to `SearchDaemon.search`.
- Normalize the response envelope (the dispatcher now tolerates both dict and list — see #4541 hardening).
- Add a servicer-level e2e test with a non-default `rrf_k` so the wire contract is exercised without mocks.

## Context

Found by adversarial review of #4541 hardening; the dispatcher-side payload is already complete (`alpha`/`fusion_method`/`rrf_k` on the wire) and envelope-tolerant, so this issue is purely the server-side RPC surface.

PR #4622, opened 2026-08-10
search: P12 adoption blockers — batch 500 + typed failures, wire contracts, index visibility, remote embedder, published plugin-host image (#4616 #4612 #4617 #4623 #4613 #4614)

### LINKED ISSUE #4612: search: gRPC proxy batch_search drops per-query typed failures — failed queries serialize as genuine empty results

### LINKED ISSUE #4613: release: post-P12, no published image provides working search — publish the plugin-host artifact + document the deployment contract

### LINKED ISSUE #4614: search-plugin: support remote/API embedding backends (post-P12 the only embedder is local mE5-small ONNX)

### LINKED ISSUE #4616: search: POST /query/batch is a hard 500 on every call post-P12 — router passes zone_id to a proxy that doesn't accept it
Found by live E2E validation of the current edge image (digest `9f7185c2`) running the full P12 topology (edge server container + real search-plugin over gRPC). Supersedes the severity of #4612 — the typed-failure regression is moot while the endpoint doesn't answer at all.

## Repro

Any `POST /api/v2/search/query/batch` with valid queries:

```
File ".../nexus/server/api/v2/routers/search.py", line 831, in search_query_batch
    await search_daemon.batch_search(fetch_specs, zone_id=zone_id)
TypeError: SearchDaemon.batch_search() got an unexpected keyword argument 'zone_id'
```

The post-P12 proxy signature is `batch_search(self, requests: list[SearchRequest])` — no `zone_id` kwarg — and it expects `SearchRequest` objects while the router passes plain spec dicts (attribute access `r.query`/`r.search_type` would fail next even if the kwarg were accepted). The endpoint was never re-wired after the pivot, and no CI covers it server-side (the e2e-edge smoke exercises permissions + build-perf, not `/query/batch`).

Single `/query` works fine against the plugin (verified: keyword/semantic/hybrid all return correct results through the full HTTP surface).

## Impact

Every batch consumer 500s. Downstream, Koodle's cross-workspace fan-out (#4583's contract, DeepBuildAI/koodle#2164) would silently demote to per-query singles via its structural fallback — losing the measured +11% coverage and paying a wasted 500 + retry per search.

## Fix sketch

Router → proxy seam needs an adapter: either the proxy accepts `(queries: list[dict], zone_id=...)` and builds `QueryRequest` protos itself (matching the pre-P12 duck type), or the router constructs `SearchRequest`s with zone stamped per entry. Plus (from #4612) mapping `sub.error` → `BatchQueryFailure` so the typed per-entry contract returns. A server-level e2e that POSTs `/query/batch` through HTTP would have caught both.

### LINKED ISSUE #4617: search: /search/index, /search/health, /search/stats response contracts drifted post-P12 (count int→object, zoneId→zone_id, health/stats fields gone)
Found by live E2E validation of the current edge image (digest `9f7185c2`) with the full P12 plugin topology, driving the Koodle SDK (strict zod parsing) against every search endpoint. Three wire-contract drifts vs the pre-P12 server (verified side-by-side against digest `63d94404` = commit 9ad238f2):

## 1. POST /api/v2/search/index — count type + key casing changed

Pre-P12: `{"status":"indexed","count":2,"zoneId":"root"}`
Post-P12: `{"status":"indexed","count":{"indexed":1,"skipped":[]},"zone_id":"root"}`

Root cause is visible in the router shim: `count = getattr(result, "indexed", result)` — the proxy now returns a **dict**, and `getattr` on a dict doesn't read keys, so the whole dict lands in `count`. `zoneId` also became `zone_id`. Downstream impact: indexing SUCCEEDS but every SDK caller's response parse fails — Koodle's indexing worker (and its search-health sentinel seeding, which gates on `count >= 1`) would throw on every successful index → retry storms against an index that already has the doc.

## 2. GET /api/v2/search/health — shape reduced

Pre-P12: `{status, initialized, daemon_initialized, backend, bm25_index_loaded, db_pool_ready, zoekt_available}`
Post-P12: `{"status":"healthy","detail":"fts + ann online"}`

## 3. GET /api/v2/search/stats — config/identity fields gone

Pre-P12: rich object incl. `initialized`, `embedding_model`, cache/consumer state.
Post-P12: `{"fts_doc_count":2,"fts_path_count":2,"ann_chunk_count":3,"parked_count":0}` only.

Health/stats consumers doing definitive reads (Koodle's search-health poller + 15-min canary page a P0 off these fields) break or go blind.

## Ask

Preserve the public wire contracts across the pivot (or version them deliberately): `count` stays an int (sum of indexed), `zoneId` casing stays, health/stats keep at least `initialized`/`embedding_model`/backend identity fields (values may honestly change — `backend: "rust-plugin"` is fine; absent keys are not). Contract tests pinning these three response shapes at the HTTP layer would keep parity honest for the rest of the migration.

### LINKED ISSUE #4618: search: keyword-lane ranking regression vs pre-P12 Postgres stack (hit@1 0.73 → 0.40 on known-item eval)
Measured during side-by-side E2E validation: identical 40-doc corpus (15 answer docs + 25 distractors) + 15 paraphrased known-item queries, path-scoped, run against the pre-P12 server (digest `63d94404`, Postgres FTS stack with page-BM25 + title arm + page aggregation, #3980/#4545) and the current edge (digest `9f7185c2`) with the real search-plugin (tantivy) over the full HTTP surface.

| lane | pre-P12 | P12 plugin | delta |
|---|---|---|---|
| keyword hit@1 | 0.733 | 0.400 | **−0.333** |
| keyword MRR@10 | 0.733 | 0.533 | −0.200 |
| hybrid hit@1 | 0.800 | 0.533 | −0.267 |
| hybrid MRR@10 | 0.900 | 0.697 | −0.203 |
| semantic MRR@10 | 1.000 | 0.883 | −0.117 (embedding model change — tracked in #4614) |

The **keyword** regression is independent of the embedding-model question: same corpus, same queries, BM25-family retrieval on both sides. The pre-P12 stack's page-level BM25 leg, title arm, and page aggregation appear to have no equivalents in the plugin's `do_keyword_query` — rare-phrase and multi-term queries that the page/title arms recovered now miss entirely. Hybrid inherits the weaker keyword arm on top of the weaker embedder, compounding to −0.20 MRR on the default lane.

Caveats: n=15, small corpus, single run — treat the deltas as directional, but they are large, deterministic (no LLM in the loop), and reproducible: happy to share the eval harness. Throughput, for balance, is dramatically better (same harness: 24 path-scoped hybrid singles at concurrency 8 — p50 439ms → 42ms, ~17 → ~92 qps), so the migration is worth it — once ranking parity for the keyword sub-fusion features lands.

Related adoption blockers from the same validation: #4616 (batch 500), #4617 (response contracts), #4614 (API embeddings), #4613 (deployable artifact).

### LINKED ISSUE #4620: search: path_contexts ranking weights (#4544 tier boost) have no effect post-P12 — CRUD works, ranking ignores the rows
Found by differential E2E probe during P12 adoption validation (edge digest `9f7185c2` + real search-plugin; pre-P12 control = digest `63d94404`).

## Repro

Seed two docs matching the same rare term under different prefixes (one keyword-stronger via repetition), query, then upsert a `path_contexts` row with `weight: 10.0` on the LOSER's prefix and re-query:

- **pre-P12:** order flips — `before=[win.md,lose.md] after=[lose.md,win.md]` ✓ (#4544 behavior)
- **P12:** order unchanged — `before=[win.md,lose.md] after=[win.md,lose.md]`, identical scores

`PUT/GET/DELETE /api/v2/path-contexts/` all work on both stacks — the rows persist; they just never reach ranking.

## Analysis

Pre-P12 the daemon's `_attach_path_contexts` loaded zone rows and applied weights post-fusion. Post-P12 the proto carries `path_prefix_boosts` on `QueryRequest` (and the plugin's query-cache keys on it), so the plugin side looks ready — but the server side never loads the `path_contexts` rows and stamps them into the request. Dead config: operators can seed weights (our deployment has ~900 rows driving search tiering, DeepBuildAI/koodle#2064) and get silent no-ops.

## Ask

Wire the router (or proxy) to resolve the zone's `path_contexts` weight rows and populate `QueryRequest.path_prefix_boosts` per query — plus an e2e that asserts a weighted row actually moves ranking through the HTTP surface, which is what would have caught this.

Sibling adoption blockers from the same validation: #4616 (batch 500), #4617 (response contracts), #4618 (stemming), #4614 (API embeddings), #4613 (artifact + shared-FS contract), #4612 (typed batch failures).

### LINKED ISSUE #4623: search-plugin: queries silently serve from a partially-built index during explicit indexing — no completion/visibility signal
Found during prod-data validation (real 75-doc corpus replayed against edge `9f7185c2` + plugin). A `POST /search/index` of a 75-doc batch takes tens of seconds (CPU embedding); during that window `type=keyword` queries return **0 hits with no error or indicator** — the FTS commit lands at the end, and an eval that queried right after the index call scored keyword hit@12 = 0.000 across the board. Re-run after settle: 1.000. The pre-P12 server made keyword results visible almost immediately after the index call for the same batch.

Impact: any write→index→query flow (Koodle's indexing worker + its search-health sentinel, which seeds a doc then requires retrieval) reads healthy-empty during builds — indistinguishable from the silent-degradation modes those monitors exist to catch.

Ask (any of): make `/search/index` return only after FTS+ANN visibility (document the latency), or expose build state via `/search/stats` (e.g. pending/committed counts per zone — `parked_count` exists, an `indexing_in_progress`/pending counter would do), or commit FTS incrementally per batch as the previous generation did. Sibling issues from the same validation: #4616, #4617, #4618, #4620.

windoliver:
Closes #4612. Closes #4613. Closes #4614. Closes #4616. Closes #4617. Closes #4618. Closes #4620. Closes #4623.

> **Also includes #4618** (folded from PR #4625, closed): `chunk_text` now indexes through tantivy's built-in `en_stem` analyzer (query parser resolves the same analyzer from the field config, so index and query time agree) — fixes the keyword-lane hit@1 0.733 → 0.400 regression from missing stemming. Index-schema change ⇒ FTS dir bumped `fts/` → `fts-v2/` (ann-`<tag>`-v2 convention, old dirs untouched) and `index_state` version bumped 1 → 2 with reset-to-empty on mismatch — without that reset the preserved mtime cache reports `Unchanged` for every file and the fresh index stays silently empty. **Operational note: first Refresh after deploy re-indexes each zone's corpus (FTS + ANN re-embed), one-time cost.** TDD: morphological probe test uses the issue's table verbatim (`authorization`/`engineers`/`batteries` + exact-form controls), watched failing pre-fix.

> **Includes #4620 in full** — PR #4626 (`fix/4620-path-context-boosts`) was folded into this PR and closed (one PR per related batch); its commits form the base of this branch. Summary of that work, reviewable as the first two commits + the final batch-lane commit:
>
> - **Single-zone lane**: `SearchRequest.path_prefix_boosts` + `_resolve_path_prefix_boosts` in the search router (per-event-loop `PathContextCache` over the path-contexts store, fingerprint-checked, fail-open) — the server finally stamps the zone's `path_contexts` weight rows onto `QueryRequest.path_prefix_boosts`. Keys are wrapped `/{prefix}/` for slash-boundary safety; pre-P12 shadowing parity preserved (weightless rows ride along as explicit 1.0 whenever any effective weight ≠ 1.0; weight-free zones send `{}` keeping the plugin's no-boost fast path).
> - **Federated lane**: `FederatedSearchDispatcher` takes an async `path_prefix_boosts_resolver(zone_id)`; each local leg carries its own zone's rows, fail-open per leg. Remote legs stay unboosted (#4556 drops all params on that surface anyway).
> - **Batch lane** (final commit here): batch is single-zone, so weights resolve once and stamp onto every inner `SearchRequest`; the shared `_request_to_pb` carries them to the wire, so Query and BatchQuery can never drift. Wiring test mirrors `test_search_prefix_boosts_wiring`.
> - **File-exact parity (Rust)**: `scoring.rs` — a boost key ending `/` also matches the path equal to the key minus that slash, so `/{p}/` covers subtree + exact-file (pre-P12 `path == prefix` behavior) with zero boundary leak (pinned by test).
> - **Live e2e** — the test #4620 asked for: `test_live_path_context_weight_moves_ranking` drives the issue's differential probe through the real HTTP surface (real kernel + real Rust plugin): `weight: 10.0` on the loser's prefix flips ranking; deleting the row restores raw BM25 order. Verified RED against develop's router (`before=[win,lose] after=[win,lose]`, the exact signature from the issue) and GREEN with the fix. Made possible by a `live_search_app` fixture upgrade: detach-signs the worktree plugin dylib with an ephemeral Ed25519 key (trusted via the loader's `NEXUS_LOCAL_TRUSTED_KEYS_DIR`), loads it via `NEXUS_PLUGIN_DIR`, points `NEXUS_SEARCH_PLUGIN_TARGET` at the spawned kernel's listener, and nulls the ReBAC search enforcer that silently emptied every response for the unseeded test key. CI behavior unchanged (suite still skips without a worktree binary).

One PR for the downstream (Koodle) P12-adoption blockers — epic #4624's **Lane A** (#4616 → #4612, "same PR is fine" per the epic's dependency map), **Lane B** (#4617), **Lane D** (#4614), **Lane E** (#4623), and **Lane F pipeline work** (#4613). Commits are one-per-concern with disjoint-or-small file overlaps — reviewable independently. Replaces #4615, #4619, #4621 (closed in favor of this). With #4620 and #4618 folded in (see notes above), this single PR now covers all 8 of the epic's blockers.

**Live-validated on a full stack built from this branch** (`nexus up --build` + real plugin host + real mE5 embedder): `scripts/test_build_perf_e2e.py` = **30/30**, including the new HTTP contract + batch-seam section.

---

## 1. `fix(search)`: batch_search proxy — hard-500 seam repair + typed per-query failures + full param forwarding (#4616, #4612)

**#4616 (the hard 500)** is defect 1 below — found here independently before the issue was filed, fixed in this commit. Its ask for "a server-level e2e that POSTs `/query/batch` through HTTP" is the new section 9 of `scripts/test_build_perf_e2e.py` (separate commit): valid query ⇒ genuine results with no `error` key; invalid spec ⇒ additive per-entry `error`. It runs on local stacks and in the e2e-edge smoke.

The P12 proxy's `batch_search` shipped with **three** defects the mock-daemon tests couldn't see (they `AsyncMock` the daemon, so the route↔proxy call contract was never exercised):

1. **Signature drift — live `/query/batch` 500s.** The route called `batch_search(fetch_specs, zone_id=...)` (dict specs + kwarg, the deleted Python daemon's shape) while the P12 proxy takes a bare positional list of typed requests. Every live call raised `TypeError` before reaching the plugin.
2. **Dropped tuning knobs.** The batch pb mapping only forwarded `q`/`zone`/`limit`/`path`/`type` — `alpha`, `fusion_method`, `rrf_k`, `expand`, and the recency params were silently discarded.
3. **Dropped typed failures (the issue as filed).** `sub.error` was discarded, so a failed inner query serialized as `{"results": [], "total": 0}` with no `error` key — regressing the 2026-08-04 contract (*absence of `error` means the result set is genuine*) that fail-closed consumers rely on for coverage accounting. The route's `BatchQueryFailure` branch was dead code.

**Fix:** route builds `SearchRequest` objects (zone_id inside each — the single-`/query` shape); proxy maps through a shared `_request_to_pb` helper (batch can never drop a knob the single path forwards) and maps `sub.error` → `BatchQueryFailure`. Stale docstring now describes the shipped #4611 behaviour (embed-cache pre-warm, concurrency default 4, clamp 1..=16 — kept at 4, that default was deliberate).

**Tests:** forwarding assertions updated to the typed shape + new `TestBatchRealProxyRoundtrip` (route → **real** proxy → fake gRPC stub) pinning the call contract, per-entry `error` serialization, and the full knob set on the wire. 58 passed; related unit suites green. Includes the surface-coverage YAML regen (router line shift).

## 2. `feat(search-plugin)`: remote/API embedding backend (#4614)

P12 left local ONNX mE5-small as the only embedder — a hard migration wall for corpora tuned on API embeddings. New `RemoteEmbedder` (OpenAI-compatible `/v1/embeddings`):

| Env var | Meaning |
|---|---|
| `NEXUS_SEARCH_EMBED_API_URL` | endpoint; presence selects the remote backend |
| `NEXUS_SEARCH_EMBED_MODEL` | model name (required with URL) |
| `NEXUS_SEARCH_EMBED_DIM` | embedding dim (required — pins the ANN index without a boot-time probe; responses validated against it) |
| `NEXUS_SEARCH_EMBED_API_KEY` / `_TIMEOUT_SECONDS` | optional bearer token / timeout (30s) |

- **Fail-loud config**: URL set but model/dim missing errors instead of silently falling back to local ONNX — a typo must not quietly reindex into a different vector space.
- **Tag contract**: `api-<sanitized-model>-<dim>` keeps `ann-<tag>-v<n>` working across embedder swaps.
- **Blocking HTTP is safe**: every `embed_batch` site runs inside `spawn_blocking`; `reqwest::blocking` uses its own runtime thread. reqwest 0.13 + rustls (already in the tree's graph; lock grows 2 lines).
- **Always compiled** (no ONNX involved); configured remote wins over `FastEmbedder`. Response items assembled by per-item `index`; provider errors surfaced truncated; >512-input batches chunked.

**Tests:** env parsing via injectable lookup (no process-env races), tag sanitization, live localhost HTTP round-trips (plain `std::net::TcpListener`, no new dev-deps): ordered assembly + bearer auth, error-status surfacing, dim-mismatch rejection. Plugin lib suite 129 passed; fmt clean; clippy clean on changed code (the `--all-targets` failures in `tests/*_e2e.rs` `mock_mut` predate this branch).

## 3. `release`: publish nexusd-cluster plugin-host image + deployment contract docs (#4613)

Post-P12 every published tag shipped with no working search — the plugin-host image existed only as an in-CI build the public can't pull.

- **New self-contained `dockerfiles/Dockerfile.nexusd-cluster-plugins`**: cluster binary (Cargo.toml-pinned nexus-vfs rev) + in-build Ed25519-signed plugin cdylib in one image. Self-containment is what enables push-by-digest multi-arch. KEEP-IN-SYNC notes added to the two sibling Dockerfiles (untouched — PR CI keeps building them).
- **`docker-publish.yml`**: new `build-cluster-platform` (amd64+arm64, native runners) + `push-cluster-manifest` jobs publish `ghcr.io/nexi-lab/nexusd-cluster` with the **same tag rules** as the server image — every server tag gets a same-commit plugin-host tag, pinnable by digest.
- **e2e-edge now runs the published image** instead of an in-workflow build — the smoke validates the exact artifact the public pulls (issue ask 3; full "push stable/latest only after smoke" gating left as noted follow-up).
- **`docs/deployment/search-plugin.md`** (+ nav): processes + image pairing, `NEXUS_SEARCH_PLUGIN_TARGET` wiring + 5s fail-soft probe, shared-filesystem requirement, signing, model provisioning (`NEXUS_SEARCH_MODEL_DIR`, `ORT_DYLIB_PATH`), remote-embedder option, pre-P12 corpus migration story (full reindex either way; remote embedder preserves the tuned vector space).

**Validation:** actionlint + `docker build --check` clean. The publish workflow only fires on develop/tag pushes, so first develop push post-merge is the live run; the cluster jobs are additive to the existing server-image path.

Also from the live validation (commit `020a9bc38`): compose templates gained the `NEXUS_SEARCH_PLUGIN_TARGET` passthrough (post-P12 a `nexus up` stack had no way to wire a plugin host without hand-editing generated compose), and `examples/standalone_host.rs` makes the full-stack validation reproducible (real plugin + real embedder over tonic; local tooling only, never a cdylib).

⚠️ First run publishes a new GHCR package `nexusd-cluster` — may need visibility set to public in org settings.

## 4. `fix(search)`: restore /index, /health, /stats wire contracts (#4617)

All three response shapes drifted across the pivot and broke strict SDK parsers / health pollers (verified live on this branch's stack **before** fixing):

- **`POST /search/index`**: the proxy's raw dict leaked whole into `count` (`getattr` on a dict reads no keys) and `zoneId` became `zone_id` — every SUCCESSFUL index threw in strict parsers (retry storms against an index that already had the doc). Now: `{"status": "indexed", "count": <int>, "zoneId": ...}`, with the route normalizing dict / `ExplicitIndexResult` / bare-int daemon returns.
- **`GET /search/health`**: restored the seven pre-pivot keys with honest post-P12 values (`backend: "rust-plugin"`, `bm25_index_loaded` from plugin status — degraded still answers keyword, `db_pool_ready` from the server's session factory, `zoekt_available: false`). A plugin dying after the boot probe now reads `status: "unavailable"` instead of a 500.
- **`GET /search/stats`**: `initialized`, `backend`, `embedding_model` restored. The model tag rides new **additive** Stats proto fields, reported from CONFIG when the embedder hasn't lazily loaded — stats polls never pay the ONNX session build. Empty ⇒ `null` (pre-P12 "no embedding model configured" semantics).

Contract tests pin all three shapes at the HTTP layer through the **real** proxy (fake gRPC stub only): `tests/integration/services/test_search_response_contracts.py` — healthy/degraded/disabled/unreachable health states, int-count + `zoneId`, stats identity + visibility fields.

## 5. `fix(search-plugin)`: index build visibility (#4623)

- **Incremental FTS commits**: `do_index_documents` commits FTS (and drops the zone's cached results) every 8 successfully indexed docs, so keyword hits appear progressively while the embed-heavy remainder of a batch is still building — pre-fix, a 75-doc batch answered healthy-empty for tens of seconds, indistinguishable from the silent-degradation modes health sentinels exist to catch. Commits cost a couple ms against tens of seconds of CPU embedding; ANN stays end-commit.
- **`indexing_in_progress`** on Stats (RAII-counted in-flight Index/IndexDocuments/Refresh ops), surfaced through `/search/stats` — pollers can tell "genuinely empty" from "still building" (the issue's option b).

`tests/index_visibility_e2e.rs` proves it with a gated embedder: keyword hits for early docs visible **while** the batch is parked mid-flight with `indexing_in_progress == 1`; counter drops to 0 and read-your-writes holds on return.

## Live validation

Full stack built from this branch (`nexus up --build`, worktree source) + real search-plugin host built from this branch (`examples/standalone_host`, real mE5-small ONNX): **`test_build_perf_e2e.py` 30/30** — HERB hybrid gate 7/8, permission-filtered search, RPC latency p50=10ms, and the new section 9 pinning the batch seam + all three #4617 contracts against live HTTP. Section 9 self-seeds the plugin via `/search/index`, so it holds on any topology (the demo pipeline seeds only the legacy engine — that CLI/legacy split is the remaining known gap, noted for follow-up)." 2>&1 | tail -1


windoliver:
## Live full-stack validation: 26/26 ✅

Ran the complete validation the batch contract deserves: `nexus up --build` from this branch (worktree source → local image, verified `_request_to_pb` inside the container), a **real search-plugin host built from this branch** (`examples/standalone_host`, real mE5-small ONNX embedder — exercises the #4614-refactored `build_default_embedder` local path), and the full `scripts/test_build_perf_e2e.py`:

```
RESULTS: 26 passed, 0 failed
```

Highlights:
- **Section 6 (HERB hybrid gate)**: 7/8 known-item hit rate through the live pipeline.
- **Section 9 (new, #4616/#4612)**: `POST /api/v2/search/query/batch` through the full HTTP surface — 200, valid entry returns genuine results (`prod-001` rank 1 via router → typed proxy → gRPC → plugin → semantic lane), invalid spec carries the additive per-entry `error`. This exact request was a hard 500 pre-fix.
- **#4614 fail-loud verified on the real binary**: partial remote-embedder config (`_API_URL` without `_MODEL`) exits with the actionable typed error instead of silently falling back to local ONNX.

Two findings from the run, folded into the last commit (`020a9bc38`):
- Compose templates gained the `NEXUS_SEARCH_PLUGIN_TARGET` passthrough — post-P12 a `nexus up` stack had **no way** to wire a plugin host without hand-editing generated compose.
- `examples/standalone_host.rs` committed so this validation is reproducible (directory-backed KernelHandle, real embedder; local tooling only, never a cdylib).

Two pre-existing gaps observed but **out of scope here** (they're epic #4624's other lanes):
- `nexus demo init` / `nexus search index` populate the legacy Postgres search path, not the P12 plugin — the plugin index stays empty until an explicit `POST /api/v2/search/index`. Filed under the #4623 index-visibility umbrella's territory.
- The explicit-index route serializes `count` as the proxy's raw dict (`{"indexed": 22, "skipped": []}`) instead of an int — exactly the #4617 response-contract complaint.

PR #4625, opened 2026-08-10
search-plugin: stem chunk_text with en_stem + fts-v2 index bump (#4618)

### LINKED ISSUE #4618: search: keyword-lane ranking regression vs pre-P12 Postgres stack (hit@1 0.73 → 0.40 on known-item eval)

windoliver:
## Problem

#4618: keyword-lane hit@1 dropped 0.733 → 0.400 vs the pre-P12 Postgres stack. Root cause (isolated in the issue): the plugin's tantivy index tokenizes `chunk_text` with the default chain — lowercase + split, **no stemmer** — so morphological query variants return zero hits (`authorization` misses "authorized", `engineers` misses "engineering", `batteries` misses "battery"). Postgres FTS stemmed both sides.

## Fix

- **`fts_index.rs`** — `chunk_text` now indexes through tantivy's built-in `en_stem` analyzer (simple tokenizer + lowercase + English Porter stemmer, pre-registered in the default `TokenizerManager`). `QueryParser` resolves the analyzer from the field config, so index and query time agree; `IndexRecordOption::WithFreqsAndPositions` kept (same as the old default TEXT), so phrase queries still work.
- **`index_manager.rs`** — index-schema change ⇒ FTS dir bumped `fts/` → `fts-v2/`, same convention as `ann-<tag>-v2`. Pre-stemmer dirs stay on disk untouched; opening the old dir with the new schema would error anyway.
- **`index_state.rs`** — `STATE_VERSION` 1 → 2, and a mismatched on-disk version now loads as **empty** instead of erroring. Without this, the preserved mtime cache reports `Unchanged` for every file and the fresh `fts-v2` index stays silently empty forever. Empty cache ⇒ next Refresh re-indexes the whole corpus into the new layout — the rebuild the issue calls for, with no operator action.

## Tests (TDD — all three watched fail first)

- `morphological_query_variants_match_stemmed_corpus_terms` — the issue's probe table verbatim (`authorization`/`engineers`/`batteries` + exact-form controls). Failed pre-fix with `query "authorization": expected 1 hit, got []`.
- `fts_dir_versioned_v2` — layout contract.
- `stale_on_disk_version_loads_as_empty_not_error` — the cache-invalidation coupling.

`cargo test -p nexus-search-plugin`: 123 lib + 32 e2e tests pass. `cargo fmt --check` clean. Clippy findings in `ann_index.rs`/e2e mocks are pre-existing and untouched.

## Operational note

First Refresh after deploy re-indexes each zone's full corpus (FTS + ANN re-embed) — one-time cost, same as any drop-and-rebuild. Old `fts/` dirs are left on disk and can be GC'd manually.

Closes #4618

### DIFF
--- a/rust/services/search-plugin/src/fts_index.rs
+++ b/rust/services/search-plugin/src/fts_index.rs
@@ -4,7 +4,7 @@
 //! # Storage layout
 //!
 //! Each zone has its own tantivy directory rooted at
-//! `~/.nexus/plugins/search/<zone_id>/fts/`.  The directory is
+//! `~/.nexus/plugins/search/<zone_id>/fts-v2/`.  The directory is
 //! plugin-owned, per-node, and NOT replicated — the SSOT for
 //! "what belongs in the index" is the kernel VFS itself, and the
 //! MetaStore `search.indexed_gen` sidecar (added in Phase 5)
@@ -18,9 +18,19 @@
 //! ```text
 //! path         STRING | STORED   exact match + retrievable
 //! chunk_index  I64    | STORED   integer, 0 in P1 (one chunk per file)
-//! chunk_text   TEXT   | STORED   BM25-analysed
+//! chunk_text   TEXT(en_stem) | STORED   BM25-analysed, English-stemmed
 //! mtime_ms     I64    | STORED   millisecond-resolution mtime
 //! ```
+//!
+//! `chunk_text` runs through tantivy's built-in `en_stem` analyzer
+//! (simple tokenizer + lowercase + English Porter stemmer) so
+//! morphological query variants match — `authorization` finds
+//! "authorized", matching the pre-P12 Postgres FTS behaviour that
+//! stemmed both sides (#4618).  The query parser resolves the same
+//! analyzer from the field config, so index and query time agree.
+//! This is an index-schema property: indices built with the
+//! pre-stemmer schema live under the old `fts/` directory and the
+//! manager opens `fts-v2/` instead — rebuild to populate.
 //!
 //! `mtime_ms` is STORED only in P1 (no INDEXED).  P6 flips it to
 //! `STORED | INDEXED` when recency queries land; INDEXED today would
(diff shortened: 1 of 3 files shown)


windoliver:
Folded into #4622 (commit 082449109 on its branch) per one-PR consolidation — same change verbatim: en_stem on chunk_text, fts/ → fts-v2/ dir bump, index_state v1→v2 graceful reset. The operational note carries over: first Refresh after deploy re-indexes each zone's corpus into the new layout.

Issue #4544, opened 2026-07-31
search: per-prefix ranking weight via path_contexts (source-tier boost)


windoliver:
## Problem

All path prefixes rank equally, but real corpora have authority tiers — curated docs vs chat transcripts vs archived material. There is no way to express "this prefix is noisier than that one" to the ranker.

The infrastructure for per-prefix metadata already exists: `path_contexts` (`src/nexus/storage/models/path_context.py`) provides longest-prefix-matched, per-zone LRU-cached, admin-CRUD-managed records, and `_attach_path_contexts` (`src/nexus/bricks/search/daemon.py:1647`, called at `:1878` / `:1918`) already walks every result set — today only to attach a `context` string.

(Refs pinned at `bd52314f7`.)

## Proposal

- Add a nullable `weight` column (default 1.0) to `path_contexts` (additive alembic migration) + expose it through the existing `/api/v2/path-contexts` CRUD.
- In `_attach_path_contexts`, multiply `result.score` by the matched prefix weight; stamp a `tier_boost` attribution field when ≠ 1.0.
- Ordering constraint: the multiply must happen **before** the final limit trim — today attachment runs after slicing, where a boost can no longer promote anything into the returned set.
- Recommended guard: skip boosting results scoring far below the top result (floor-ratio gate) so metadata can reorder near-peers but can't lift weak matches past strong ones.

## Acceptance

- Seeding `weight=0.5` on a noisy prefix demotes its hits below equal-relevance hits from weight-1.0 prefixes; `weight` unset/1.0 everywhere is byte-identical to today.
- Boost visible per-result; federated path applies the owning zone's weights.

## Dependencies

Soft: fusion-params issue (knob plumbing pattern). Hard: none.

Issue #17, opened 2025-10-12
[v0.2.0] Design and implement parser system architecture


windoliver:
## Description
Create extensible parser system for processing various document formats.

## Architecture Reference
See NEXUS_COMPREHENSIVE_ARCHITECTURE.md Section: "Parser System Architecture" and "Rich Format Parsing"

### Parser Interface
```python
class Parser(ABC):
    @abstractmethod
    def can_parse(self, file_path: str, mime_type: str) -> bool:
        """Check if this parser can handle the file"""

    @abstractmethod
    async def parse(self, content: bytes, metadata: dict) -> ParseResult:
        """Parse file content and return structured data"""

    @property
    @abstractmethod
    def supported_formats(self) -> List[str]:
        """List of supported file extensions"""
```

### ParseResult Schema
```python
class ParseResult:
    text: str                    # Extracted text
    metadata: dict              # File metadata
    structure: dict             # Document structure
    chunks: List[TextChunk]     # Semantic chunks
    images: List[ImageData]     # Extracted images
```

### Core Parser System
- [ ] Create `Parser` abstract base class
- [ ] Create `ParseResult` data class
- [ ] Implement `ParserRegistry` for managing parsers
- [ ] Add parser selection logic (by extension and MIME type)

### Parser Registry
- [ ] Auto-discover parsers on initialization
- [ ] Register parsers by format
- [ ] Handle multiple parsers for same format (priority)
- [ ] Provide `get_parser(file_path)` method

### Document Type Detection
- [ ] Use `python-magic` for MIME type detection
- [ ] Fallback to extension-based detection
- [ ] Handle compressed files (.gz, .zip)
- [ ] Handle text encoding detection

### Error Handling
- [ ] Define `ParserError` exception
- [ ] Graceful degradation (extract what's possible)
- [ ] Logging for failed parsing
- [ ] Return partial results when possible

### Testing
- [ ] Test parser registration
- [ ] Test parser selection logic
- [ ] Test error handling
- [ ] Test with various file types

## Acceptance Criteria
- ✅ Extensible parser system implemented
- ✅ Parser registry works
- ✅ Document type detection accurate
- ✅ Error handling robust
- ✅ Tests pass with >80% coverage

## Related Files
- `src/nexus/parsers/base.py`
- `src/nexus/parsers/registry.py`
- `src/nexus/parsers/types.py`
- `tests/unit/test_parser_system.py`

PR #4626, opened 2026-08-10
search: wire path_contexts weights into QueryRequest.path_prefix_boosts (#4620)

### LINKED ISSUE #4620: search: path_contexts ranking weights (#4544 tier boost) have no effect post-P12 — CRUD works, ranking ignores the rows

windoliver:
## Problem

#4620: `path_contexts` ranking weights (#4544 tier boost) are dead config post-P12. The plugin fully honours `QueryRequest.path_prefix_boosts` (applies the longest `starts_with`-matching multiplier post-fusion, keys its query cache on the map) — but the server never loaded the zone's rows into the request. CRUD persisted rows; ranking silently ignored them.

## Fix (commit 1 — single-zone lane)

- **`contracts/search_types.py`** — `SearchRequest.path_prefix_boosts: dict[str, float] | None`, keys already in the plugin's wire shape.
- **`bricks/search/path_context.py`** — `prefix_boosts_from_records`: stored prefixes are slash-free canonical, plugin hit paths are slash-prefixed and matched with plain `starts_with`, so keys are rewrapped as `/{prefix}/` (directory-boundary safe — `docs` can't boost `/docs-v2/…`; `""` stays the zone-wide key). Pre-P12 parity: the longest matching *record* won even weightless (multiplier defaulted 1.0), shadowing shorter weighted prefixes — so when any row has an effective weight ≠ 1.0, all rows ride along (weightless as explicit 1.0); weight-free zones return `{}` to keep the plugin's no-boost fast path.
- **`routers/search.py`** — `_resolve_path_prefix_boosts`: per-event-loop `PathContextCache` over the path-contexts store (fingerprint-checked → an upsert is visible on the next query, no full table scan per request). Fail-open to unboosted search; warns once per process. Stamped onto the single-zone `SearchRequest`.
- **`bricks/search/daemon.py`** — the proxy stamps the field onto `QueryRequest` for both `Query` and `BatchQuery`.

## Fix (commit 2 — the flagged gaps)

- **Federated lane**: `FederatedSearchDispatcher` takes an optional async `path_prefix_boosts_resolver(zone_id)`; each **local** leg now carries its own zone's weight rows, resolved through the same per-loop cache. Fail-open per leg (a resolver error costs the boost, never the leg). Remote legs stay unboosted — the remote RPC surface drops every search param today (#4556) and remote rows belong to the remote deployment. Dispatcher is per-request, so its result cache can't serve stale boosts.
- **File-exact prefix parity (Rust)**: `scoring.rs::apply_prefix_boost` — a key ending in `/` now also matches the path equal to the key minus that slash. `/{p}/` therefore covers both the subtree and the exact-file case (pre-P12 matched via `path == prefix`), still with zero boundary leak (`/docs/` cannot touch `/docs-v2/…`, pinned by test).

## Tests (TDD — every RED watched first)

- Unit: `prefix_boosts_from_records` translation rules (5).
- Wiring: `/api/v2/search/query` against a real sqlite store + capturing daemon — weighted row reaches the daemon as `{"/docs/": 5.0}`, zone isolation, staleness refresh, fail-open without a store, **federated legs carry per-zone boosts** (6).
- Daemon proto mapping (2). Dispatcher resolver behavior incl. fail-open (3). Rust: exact-file match + boundary-leak pin (2; 122/122 plugin lib tests green).
- **Live e2e — the test the issue asked for**: `test_live_path_context_weight_moves_ranking` runs the issue's differential probe through the real HTTP surface (real kernel + real Rust plugin): `weight: 10.0` on the loser's prefix flips the order; deleting the row restores raw BM25 order. Verified RED against develop's router — `before=[win,lose] after=[win,lose]`, the exact signature from the issue — and GREEN with this fix.

## Live-e2e fixture upgrade (what made the HTTP-surface test possible)

The `live_search_app` fixture previously always 503'd on `/search/*`: the SearchDaemon dials `127.0.0.1:2126` but the locally spawned kernel binds a random port, and no plugin was ever loaded. The fixture now (all opt-in, only when a worktree `libnexus_search_plugin.dylib/.so` exists):
- detach-signs the dylib with an ephemeral Ed25519 key and trusts it via the loader's `NEXUS_LOCAL_TRUSTED_KEYS_DIR` extension point (kernel refuses unsigned plugins; format matches `scripts/sign_plugin.py`),
- loads it via `NEXUS_PLUGIN_DIR` before `nexus.connect`,
- points `NEXUS_SEARCH_PLUGIN_TARGET` at the spawned kernel's listener (the plugin's SearchService rides the kernel's own gRPC server via Phase-P routing),
- nulls `app.state.permission_enforcer` — Decision-#17 ReBAC denied 100% of hits for the unseeded test API key, silently emptying every `/search/query` response (`NEXUS_ENFORCE_PERMISSIONS` only governs VFS ops).

CI behavior unchanged: no worktree binary → suite skips exactly as before. `plugin_loaded=False` → the new test skips with a build hint.

## Remaining scope notes

- **Batch endpoint**: the daemon proxy forwards boosts for `BatchQuery`; the batch *route* stamping lands on PR #4622 (which owns the batch-route rework and is now stacked on this PR).
- Pre-existing test rot surfaced, not fixed here: `tests/integration/services/test_search_router.py::TestSearchQueryEndpoint` and most of `tests/integration/bricks/search/test_federated_search.py` fail identically on develop (mocks predate the P12 positional-`SearchRequest` daemon call — verified by running both suites against pristine develop).
- Merge-conflict heads-up: touches `daemon.py` + `search.py`, same files as PR #4622 — #4622 is now rebased onto this branch, so land this first.

Closes #4620

### DIFF
--- a/rust/services/search-plugin/src/scoring.rs
+++ b/rust/services/search-plugin/src/scoring.rs
@@ -157,6 +157,14 @@ pub fn apply_recency(
 /// multiply the score by that entry's multiplier.  No matching
 /// prefix ⇒ score unchanged (multiplier 1.0).
 ///
+/// A key ending in `/` also matches the path that EQUALS the key
+/// minus that trailing slash (#4620): the server wraps stored
+/// path-context prefixes as `/{p}/` for slash-boundary safety, and a
+/// row whose prefix names an exact file (`/notes/hello.md/` after
+/// wrapping) must still boost that file — the pre-P12 stack matched
+/// it via `path == prefix` equality.  Plain `starts_with` keys keep
+/// their raw semantics.
+///
 /// Longest-match semantics matter when boosts overlap
 /// (`/docs/` = 1.5, `/docs/api/` = 2.0) — the more specific
 /// prefix wins.  Ties (same length) break arbitrarily but stably
@@ -168,7 +176,10 @@ pub fn apply_prefix_boost(hits: &mut [QueryResult], boosts: &HashMap<String, f32
     for hit in hits.iter_mut() {
         let best = boosts
             .iter()
-            .filter(|(prefix, _)| hit.path.starts_with(prefix.as_str()))
+            .filter(|(prefix, _)| {
+                hit.path.starts_with(prefix.as_str())
+                    || (prefix.ends_with('/') && hit.path == prefix[..prefix.len() - 1])
+            })
             .max_by_key(|(prefix, _)| prefix.len());
         if let Some((_, &multiplier)) = best {
             hit.score *= multiplier;
@@ -347,6 +358,39 @@ mod tests {
     }
 
     // ── apply_prefix_boost ─────────────────────────────────────
+
+    #[test]
+    fn prefix_boost_trailing_slash_key_matches_exact_file_path() {
+        // #4620 parity: the server wraps stored path_contexts prefixes
+        // as "/{p}/" for slash-boundary safety.  A row whose prefix
+        // names an exact FILE ("/notes/hello.md/" after wrapping) must
+        // still boost that file — pre-P12 matched it via
+        // `path == prefix` equality.
+        let mut hits = vec![hit("/notes/hello.md", 1.0, None)];
+        let mut boosts = HashMap::new();
+        boosts.insert("/notes/hello.md/".to_string(), 3.0);
+        apply_prefix_boost(&mut hits, &boosts);
+        assert!(
+            (hits[0].score - 3.0).abs() < 1e-6,
+            "exact-file trailing-slash key must boost, got {}",
+            hits[0].score
+        );
+    }
+
+    #[test]
+    fn prefix_boost_trailing_slash_key_does_not_leak_to_siblings() {
+        // The boundary guarantee that motivated the "/{p}/" wrapping:
+        // "/docs/" must not touch "/docs-v2/…" or "/docsfile.md".
+        let mut hits = vec![
+            hit("/docs-v2/x.md", 1.0, None),
+            hit("/docsfile.md", 1.0, None),
+        ];
+        let mut boosts = HashMap::new();
+        boosts.insert("/docs/".to_string(), 5.0);
+        apply_prefix_boost(&mut hits, &boosts);
+        assert!((hits[0].score - 1.0).abs() < 1e-6, "sibling dir leaked");
+        assert!((hits[1].score - 1.0).abs() < 1e-6, "sibling file leaked");
+    }
 
     #[test]
     fn prefix_boost_empty_map_is_noop() {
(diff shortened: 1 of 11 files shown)


windoliver:
CI triage from the stacked PR #4622 (which bases on this branch): both current failures reproduce here and have mechanical fixes —

1. **search-plugin E2E**: `AttributeError: '_MinimalSearchRequest' object has no attribute 'path_prefix_boosts'` — the daemon proxy now reads `request.path_prefix_boosts` unconditionally, but the duck-typed stand-in in `tests/integration/bricks/search/test_search_daemon_live_plugin.py` doesn't carry the field. Fix (already applied on the stacked branch as `468930dcd`, copy verbatim):

```python
# _MinimalSearchRequest.__init__ signature:
path_prefix_boosts: dict[str, float] | None = None,
# body:
self.path_prefix_boosts = path_prefix_boosts
```

2. **Enforce API/RPC Surface Matrix**: the freshness gate — this branch's router hunks shift route line numbers, so regenerate and commit:

```bash
uv run python scripts/gen_api_surface_coverage.py
git add docs/surface-coverage/api-rpc-surface-coverage.yaml
```

Since #4622 is stacked on this PR, these two block that whole train — happy to re-stack as soon as they land here.

windoliver:
Folded into #4622 per one-PR-per-batch: this branch (fix/4620-path-context-boosts) is the base of #4622's stack, so every commit here — single-zone + federated boost wiring, daemon stamping, Rust exact-file trailing-slash match, live-e2e fixture upgrade + ranking-flip e2e — lands through #4622, whose final commit adds the batch lane. Branch left in place until #4622 merges (it is the stack base).

Issue #4541, opened 2026-07-31
search: alpha/fusion request params are validated then ignored on the backend path; rrf_k not exposed


windoliver:
## Problem

`GET /api/v2/search/query` accepts and validates `alpha` (0.0–1.0) and `fusion` (`rrf|weighted|rrf_weighted`), and `SearchDaemon.search()` threads them through (`src/nexus/bricks/search/daemon.py:1784-1791`) — but the primary indexed path `_search_via_backends` (`daemon.py:2012`) takes neither. Hybrid always runs the hardcoded two-stage fusion:

```
kw_fused = rrf_fusion(chunk_kw, page_kw, k=60, limit=limit * 2, id_key=None)   # daemon.py:2128
fused    = rrf_fusion(kw_fused, dense,   k=60, limit=limit,     id_key=None)   # daemon.py:2129
```

Only the legacy fallback `_hybrid_search` receives `fusion_method`, and it marks the param `# noqa: ARG002` (`daemon.py:2652`). Net effect: a caller can send fusion knobs, get a 200, and nothing changes. `rrf_k` is not exposed at all, even though `fusion.py`'s `FusionConfig` supports method/alpha/k/top_rank_bonus.

(Refs pinned at `bd52314f7`.)

## Proposal

- Thread `alpha` + `fusion_method` (+ a new optional `rrf_k`, default 60) from the route into `_search_via_backends`.
- Use `fuse_results(...)` / `rrf_weighted_fusion(...)` for the `kw_fused × dense` stage according to the requested method; the page-BM25 sub-fusion can stay plain RRF.
- Defaults must stay byte-identical (`rrf`, k=60, alpha ignored under plain RRF) so existing callers see no change.

Related but out of scope: the `rerank` request param is likewise accepted-and-ignored (`# noqa: ARG001` in `routers/search.py`) — there is no reranker stage to wire since the txtai drop. It should be documented as inert (or rejected) until a reranker exists.

## Acceptance

- `fusion=weighted&alpha=0.9` produces a different ordering than the default on a fixture corpus; `fusion=rrf_weighted&alpha=0.0` ranks by keyword legs only.
- Default-param request is byte-identical to today (regression test).
- Test asserting the route params actually reach the fusion call.

## Dependencies

Independent. Recommended **first** of the ranking series — the later boosts (recency, tier, title arm) each want per-request knobs and should reuse this plumbing.

windoliver:
[...]

PR #4635, opened 2026-08-10
feat(search-plugin): mirror title-arm hybrid fusion into the Rust plugin (#4628)

### LINKED ISSUE #4628: search-plugin: mirror title-arm hybrid fusion (#4552) — the last unmirrored windoliver search feature
## Context

Windoliver's Python search feature cluster from late July/early August landed 10 PRs before P12 (#4598) deleted the Python daemon.  Nine of them have equivalents in the Rust ``nexus-search-plugin``:

| Python PR | Feature | Rust plugin status |
|---|---|---|
| #4550 | page_aggregation on final fused list | ✅ ``fusion::pool_by_document`` in ``service.rs`` L831 + L1720 |
| #4551 | ``alpha``/``fusion``/``rrf_k`` on hybrid path | ✅ wired proto → ``FusionOpts`` in service |
| #4553 | recency decay | ✅ ``scoring::apply_recency`` (already mirrored per project memory) |
| #4554 | per-prefix tier boost via path_contexts | ✅ ``scoring::apply_prefix_boost`` (``//! Matches Python #4544``) |
| #4558-#4560 | fusion-param hardening + read-gate adversarial closes | ✅ folded into above |
| #4567 | bounded ``file_paths`` projection wait | N/A — plugin owns the index, no projection lag |
| #4583 | ``/query/batch`` parity — concurrent, embed-once | ✅ #4610/#4611 (parallel BatchQuery + embed cache) |
| #4586 | keyword-only degradation on batch | ✅ #4611 covers this shape |

**Missing:** #4552 — skeleton title arm folded into hybrid fusion as a third RRF input.

## The gap

Python's ``SearchDaemon.locate()`` returns skeleton-index hits (an in-memory BM25-lite over path tokens + titles).  #4552 folded those hits into ``_search_via_backends`` as a THIRD keyword-side arm — the sub-fusion became ``rrf_multi_fusion([chunk, page, title])``.  Title-shaped queries (``<project> design doc``) ranked their target even when body chunks were weak.

The Rust plugin has ``Locate`` as an RPC (path→indexed check) but not as a third fusion input in the ``Query`` path.  A caller who fires ``<project> design doc`` today gets the same weak result the Python daemon got PRE-#4552.

## Ask

- Add a title-arm run (per-zone skeleton index over path segments + first-heading text — the plugin already has both from indexing time) that produces a ranked list.
- In ``do_semantic_query`` / ``do_hybrid_query``, extend the sub-fusion so the keyword lane fuses three arms (chunk-FTS, page-pool, title) via ``rrf_multi_fusion`` before the final alpha/fusion stage.
- Gate on a per-request or env knob mirroring ``NEXUS_SEARCH_TITLE_ARM=true`` (Python default), attribute per-hit ``title_score`` for observability.
- E2E: title-shaped query surfaces the intended file even when body chunks are weak (existing plan doc has the workload shape).

Design SSOT: ``docs/superpowers/specs/2026-07-31-title-arm-hybrid-fusion-design.md``.

## Non-goals

- Reindexing.  Skeleton signal is derivable from existing indexed data (path + first heading).
- Cross-zone.  Same single-zone scope as the Python daemon; federation dispatch stays kernel-tier.

windoliver:
Closes #4628 — the last unmirrored windoliver search feature (Python #4552).

## What

- **Per-zone in-memory skeleton** (path tokens + first-heading title) derived from already-indexed FTS data — no schema change, no reindex (issue non-goal honored).
- **`ZoneSkeleton::locate()`**: port of the deleted Python `SearchDaemon.locate()` — title×2 + path×1 overlap scoring, per-field DF caps (1024), oversized-bucket prefix rescue (200k scan budget), aggregate token/candidate budgets (12/4096), stopword stripping with exact-title fallback, deterministic (-score, path) ordering. Constants copied verbatim from the pre-P12 daemon.
- **Generation-keyed skeleton cache** on `IndexManager`: FTS commits bump the tantivy searcher generation, so staleness detection is structural — no invalidation call sites to maintain, and new mutation paths can't forget to invalidate.
- **Hybrid keyword lane**: `fusion::rrf_multi([chunk, title])` sub-fusion with Python's top-rank bonus (+0.05/+0.02), BEFORE the untouched #4541 alpha/fusion final stage. **Empty title arm ⇒ byte-identical pass-through** — non-title queries are unchanged under every fusion method incl. WEIGHTED (pinned by e2e).
- **Hydration** mirrors Python: borrow the best fetched keyword-leg row per path (aligns the (path, chunk_index) fusion key so RRF votes accumulate), dense rows lowest-priority, capped (32) FTS chunk-0 fetch for uncovered paths, chunkless docs stay retrievable with empty text. Fail-soft end to end — the arm can never fail a search.
- **`QueryResult.title_score`** attribution (optional proto field) + Python shim passthrough in `_result_to_base`; the HTTP/batch/federated serializers already handle omit-when-None (that plumbing survived P12).
- **Kill-switch**: `NEXUS_SEARCH_TITLE_ARM` (default on, `false|0|no` disables), plus a builder pin for tests.

## Scope decisions

- No page arm: Python's page-BM25 leg was Postgres-only; the SQLite branch passed an empty page arm (design-SSOT no-op). The Rust plugin mirrors that shape; `rrf_multi` is N-way ready for a future page arm. Deriving a fake page arm from pooled chunk hits would double-vote chunk evidence and break non-title-query parity.
- Title extraction = first ATX heading in the 2 KiB chunk-0 head (frontmatter-skipped). Python's per-extension extractor registry lived in the catalog brick; docs are the title-shaped-query use case. Title-less docs still join on path tokens.
- Hybrid-only, single-zone — same as the Python design SSOT (`docs/superpowers/specs/2026-07-31-title-arm-hybrid-fusion-design.md`).

## Testing

- **Unit** (38 new): tokenizer camelCase/acronym parity, heading extraction + frontmatter + head-cap, locate scoring (7.0 acceptance pin), stopword/exact-title/DF-cap/prefix-rescue/determinism, rrf_multi reference-formula + attribution + key-alignment, skeleton cache generation semantics.
- **E2E** (`title_arm_e2e.rs`): title-shaped query surfaces a weak-body doc with `title_score=7.0` through the real Query RPC; arm-off shows no attribution and never-better target rank; non-title query returns **byte-identical** results arm on/off.
- Full crate suite 201/201 green; Python transport + serializer suites 16/16 green. Clippy: no new-code findings (3 pre-existing in untouched R8-era code left alone).


windoliver:
Live e2e verified on a full `nexus up --build` stack (worktree image + host-side `nexusd-cluster` plugin host running this branch's plugin):

- Title arm live through the whole stack: hybrid `atlas design doc` → `title_score=7.0` on the target (same pin as the unit/e2e suites), HERB "Nexus Core" top hit carries `title_score=4.0`.
- `scripts/test_build_perf_e2e.py`: **30 passed, 0 failed** — HERB hybrid hit rate 8/8, index-ready in 5s, RPC p50 5ms, `backend=rust-plugin` contract + batch seam green.
- Found + fixed along the way (commit 617ce4081): post-P12, the gRPC `semantic_search` dispatcher's daemon gate only sniffed pre-P12 backend attrs, so every CLI/gRPC search silently fell back to SQL ILIKE (wrong ranking, no fusion, no attribution) while HTTP worked. Gate now recognizes the plugin shim; plugin-less deployments keep the fallback. Unit tests pin both sides.

windoliver:
Adversarial review-fix loop complete: 10 rounds of Codex adversarial review, 20 findings fixed across commits 74f354563..e9feba3b6 (all suites green after each round; final: 213 Rust tests + Python gate suite).

Highlights, most severe first:
- **Epoch-stamped query cache validated on read** — entries carry the producing query's start FTS generation; any commit racing a leg, the fusion, or the insert itself makes them unservable (kills the TOCTOU + mixed-generation-leg windows; also covers the pre-existing kw/sem race).
- **Per-zone durable dirty window** — writers mark a dir-fsynced `.write-dirty` sentinel before the first sink mutation (fail-closed if it can't persist), clear it only after every sink committed AND the pass verifiably converged (no transient skips, no ANN-retryable docs, no kept tombstones, no walk-skipped subtrees); dirty zones bypass the cache entirely, and the mark survives restarts.
- **Full-root pre-dirty refresh = reconciliation mode** — distrusts the mtime cache, exempt from the repair cap so large zones can't stall in cache-bypass.
- **Skeleton hardening** — single-flight builds (Fresh/Stale/Building), 2 KiB aggregate head retention during the scan, head reassembly across chunker-sealed frontmatter/preamble, strict ATX + marker-aware fence grammar, markdown-extension-only titles.
- **Hydration safety** — single-decode representative lookups off the async runtime, FTS-liveness + exact-chunk confirmation before borrowing dense identities (no ANN-orphan resurrection), all-dropped arms leave the keyword lane byte-identical.
- **Fail-closed shim gate** — enforcing deployments return nothing on context-less gRPC searches (matches the SQL fallback posture).

Remaining known-accepted: 3 pre-existing clippy lints in untouched R8-era code.

Issue #4634, opened 2026-08-10
search-plugin: apply prefix boosts + recency inside candidate collection (replace fixed over-fetch bound)


windoliver:
Post-fusion score adjustments (path-context boosts #4620, recency #4543) act on an over-fetched pool — after #4622's hardening, 10x the requested limit clamped at 500 (10x = the path-context weight cap). That reaches a boosted doc starting ~rank 4 at limit=1 (regression-tested), but it is a pragmatic bound, not a proof: a doc scoring >10x below the unadjusted cut line is still unreachable.

Proper fix: push the boost/recency multipliers into candidate scoring/collection (tantivy collector + ANN pre-scoring), or merge separately collected candidates from boosted prefixes, then fuse once. Also fold in a refresh resume cursor (persisted continuation past the max_docs repair budget) if touching the same surfaces.

Context: #4620, #4622 review rounds R3-R5.

Issue #4633, opened 2026-08-10
release: atomic server+plugin-host pair consumption via release manifest


windoliver:
#4622 shipped smoke-gated paired promotion with fail-closed snapshots and symmetric rollback, but tag-pair consistency is still best-effort: registries offer no cross-repo transaction, so an in-window pull (or a rollback that can't delete a freshly minted tag — GHCR's version-delete removes ALL tags on the manifest incl. the staging SHA) can observe a split pair.  Digest pinning is the current hard guarantee (documented in docs/deployment/search-plugin.md).

Follow-up: publish an immutable release manifest recording both smoke-verified digests (e.g. an OCI artifact or a JSON file tagged per release) and teach deployments/compose to consume the manifest instead of independent mutable tags. Then mutable tags become human conveniences rather than the deployment contract.

Context: #4613, #4622 review rounds R2-R7.

Issue #4632, opened 2026-08-10
search-plugin: park + retry transient embed failures (remote 429/timeout) instead of relying on refresh re-pass


windoliver:
#4622's review hardening made embed failures RETRYABLE (incomplete docs record mtime None, so Refresh/IndexDocuments re-processes them) — but retries are passive: nothing schedules them, and a remote provider outage means every affected doc re-chunks + re-embeds on the next full pass.

The designed mechanism exists: the parked-event queue (`parked_state`, ParkedList RPC, `IndexDocumentsResponse.parked_paths` — currently always empty pending "step 4"). Wire embed failures into it with bounded provider-aware retries (Retry-After honored for remote embedders), and surface the queue depth on Stats next to `indexing_in_progress`.

Context: #4614, #4622 review rounds R1/R2.

Issue #4631, opened 2026-08-10
search-plugin: adopt shared KeyLockMap/OpCounter (#4629) — blocked on leaf-crate extraction


windoliver:
PR #4629 extracted `services::locks::KeyLockMap` and `services::ops::{OpCounter, OpGuard}` explicitly for downstream cdylib plugins, and search-plugin hand-rolls equivalent primitives (per-zone write locks on IndexManager, IndexingGuard from #4622's review hardening).

Blocker: the `services` crate carries an UNCONDITIONAL `kernel` workspace dependency, so depending on it from the plugin pulls the whole kernel into the cdylib graph — heavy and against the plugin's plugin-abi-only isolation posture.

Suggested: extract `locks`/`ops` into a tiny leaf crate (no kernel/tonic/tokio deps — they only need parking_lot + std atomics), have `services` re-export it, then swap search-plugin's hand-rolled versions. Mechanical after the extraction; existing tests (index_visibility_e2e, concurrent-write coverage) pin behavior.

PR #4637, opened 2026-08-10
fix(search-plugin): read-side orphan guard — drop ANN hits without FTS rows

windoliver:
## Summary

Read-side defense-in-depth for the ANN lane: `do_semantic_query` now drops ANN hits whose path has **no corresponding FTS row** when the FTS index is open and readable.

This was documented residual #3 from the PR #4622 adversarial review ("ANN-without-FTS read guard"). The write-side lifecycle work in #4622 (embedder-generation-aware `IndexState`, delete tombstones, content-transition purges, per-zone write locks) closes every *known* path that creates orphaned vectors — but any *unknown* or historical drift (crash between ANN insert and FTS commit, pre-#4622 indexes, manual index surgery) could still surface ANN hits pointing at documents that no longer exist in FTS. Those hits previously came back with empty `chunk_text` and no mtime — junk results the caller can't distinguish from real ones.

## Behavior

`enrich_ann_hit` returns `Option<QueryResult>`:

- **FTS open, no row for the hit's path** → hit dropped, `tracing::debug!` orphan-guard log. These are provably stale vectors.
- **FTS unavailable** (failed to open) → hit kept with empty text, as before. Absence of FTS is not evidence the document is gone; degrading semantic search to zero results on an FTS outage would be worse than returning unverifiable hits.

Cost is effectively zero: the FTS lookup per ANN hit was already being performed to materialize `chunk_text`/`mtime_ms` — this only changes what happens when that lookup misses.

## Testing

- New regression test `semantic_query_drops_ann_hits_without_fts_rows` in `tests/index_visibility_e2e.rs`: indexes docs, deletes one's FTS rows out from under the ANN index, asserts the orphaned path never appears in semantic results while live paths still do.
- Full plugin suites green locally: 137 lib tests + all e2e integration tests.

## Related

- Residual from #4622 review (round 10 documented residuals).
- Follow-up tracking issues filed alongside this: #4630 (CLI/legacy seeding split), #4631 (KeyLockMap/OpCounter migration), #4632 (embed-failure parking), #4633 (release manifest), #4634 (boost-inside-collection).

### DIFF
--- a/rust/services/search-plugin/src/service.rs
+++ b/rust/services/search-plugin/src/service.rs
@@ -786,8 +786,6 @@ fn do_semantic_query(
     // Materialise chunk_text + mtime via the FTS index — the ANN
     // stores only vectors + paths, but the RPC contract carries the
     // full QueryResult shape so callers don't need a follow-up read.
-    // A missing FTS row (drift scenario) leaves chunk_text empty;
-    // the path itself is still meaningful.
     let fts = manager.get_or_open(zone_id).ok();
 
     let mut out = Vec::with_capacity(limit);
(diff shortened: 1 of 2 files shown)


windoliver:
Rebased onto develop (now containing #4635's title arm) — the `service.rs` conflict resolved by combining both changes: `enrich_ann_hit` keeps #4635's `title_score: None` field inside this PR's `Some(...)` orphan-guard wrapper. Both features verified live together on a full `nexus up --build` stack:

- **Orphan guard validated end-to-end**: indexed a doc (FTS+ANN), `NotifyFileChange` delete (FTS dropped, ANN cleanup deferred), then a semantic query for its unique text — pre-delete it ranked #1, post-delete it is no longer served despite the vectors still being physically present.
- **Title arm unaffected**: hybrid `atlas design doc` still returns `title_score=7.0`; the two guards compose (this PR filters the dense leg's own results; #4635's hydration gate covers title-arm borrows).
- `scripts/test_build_perf_e2e.py`: **30 passed, 0 failed** (HERB 8/8, RPC p50 4.7ms).
- Full crate suite 214/214 green after the rebase; fmt-check clean; CI-faithful clippy (`--workspace`-style lib lint) clean.

Issue #235, opened 2025-10-23
v0.4.0 Demo: Multi-Backend Data Processing with Claude Code Agent


windoliver:
## Overview
Create an end-to-end demo showcasing an AI coding agent (Claude Code or similar) reading from multiple storage backends, processing data, and writing results back. This demonstrates Nexus's unified multi-backend abstraction and real-world AI agent integration.

**Demo Type:** Multi-backend AI agent data processing  
**Agent Count:** 1 (Claude Code or equivalent AI coding agent)  
**Complexity:** Medium-High  
**Estimated Time:** 2-3 days  

## Demo Scenario

**Cross-Cloud Data ETL:**
1. **Source 1**: Read customer data from **Google Cloud Storage** (GCS)
2. **Source 2**: Read transaction logs from **Local filesystem**
3. **Process**: AI agent (Claude Code) analyzes and merges data
4. **Destination 1**: Write summary report to **GCS**
5. **Destination 2**: Write detailed logs to **Local**

**All through unified Nexus API - no cloud SDK code in agent!**

## Features Demonstrated

- ✅ **Multi-backend unification** (GCS + Local in single namespace)
- ✅ **AI coding agent integration** (Claude Code, Cursor, etc.)
- ✅ **Transparent cloud access** (agent doesn't know it's GCS)
- ✅ **Cross-backend data processing** (read from one, write to another)
- ✅ **Content-addressable storage** (automatic deduplication)
- ✅ **Configuration-based routing** (YAML backend config)
- ✅ **Real-world ETL use case**

## Architecture

\`\`\`
┌────────────────────────────────────────────────────────────┐
│  AI Coding Agent (Claude Code)                             │
│                                                             │
│  Python Code:                                              │
│    with open('/nexus/gcs/customers.csv', 'r') as f:       │
│        data = f.read()  # Actually reading from GCS!       │
│                                                             │
│    with open('/nexus/local/processed.csv', 'w') as f:     │
│        f.write(result)  # Writing to local!                │
└────────────────┬───────────────────────────────────────────┘
                 │
                 │ Standard File I/O
                 │
┌────────────────▼───────────────────────────────────────────┐
│  Nexus Unified Filesystem                                  │
│                                                             │
│  ┌─────────────┬─────────────┬─────────────┐             │
│  │ /gcs/       │ /local/     │ /s3/        │             │
│  │ (GCS)       │ (LocalFS)   │ (Future)    │             │
[...]

Issue #229, opened 2025-10-23
v0.4.0: Firecrawl Plugin - Production-Grade Web Scraping Integration


windoliver:
## Overview
Create `nexus-plugin-firecrawl` to integrate Firecrawl's production-grade web scraping capabilities into Nexus, enabling high-quality documentation scraping for skill generation.

## Problem Statement
Current `nexus-plugin-skill-seekers` uses basic BeautifulSoup scraping which:
- ❌ Cannot handle JavaScript-rendered content
- ❌ Gets blocked by anti-bot protections
- ❌ No proxy management
- ❌ Poor quality on complex documentation sites

**Firecrawl** solves this with:
- ✅ JS rendering (Puppeteer/Playwright)
- ✅ Anti-bot detection handling
- ✅ Proxy rotation
- ✅ PDF/complex document support
- ✅ Batch processing
- ✅ LLM-ready markdown output

## Proposed Solution
Build `nexus-plugin-firecrawl` that wraps Firecrawl API and integrates with Nexus filesystem and plugin ecosystem.

### 1. Single URL Scraping
```bash
nexus firecrawl scrape https://docs.stripe.com/api
nexus firecrawl scrape https://example.com --output scraped/example.md
nexus firecrawl scrape https://example.com --json  # For piping
```

### 2. Website Crawling
```bash
nexus firecrawl crawl https://docs.stripe.com --max-pages 100
nexus firecrawl crawl https://docs.stripe.com \
  --include-paths "/api/**" \
  --exclude-paths "/blog/**"
```

### 3. Unix Pipe Mode (CRITICAL)
```bash
# Pipe to skill creation
nexus firecrawl pipe https://docs.stripe.com/api | \
  nexus skills create-from-web --name stripe-api

# Multi-stage pipeline
nexus firecrawl pipe https://docs.stripe.com/api | \
  jq '.content' | \
  nexus skills create-from-web | \
  nexus anthropic upload-skill
```

### 4. Additional Commands
- `nexus firecrawl map <url>` - Discover all URLs
- `nexus firecrawl search <query>` - Web search + scrape
- `nexus firecrawl extract <url> --schema schema.json` - Structured extraction

### Phase 1: Core Plugin Structure
- [ ] Create plugin package
- [ ] Setup entry points
- [ ] Configuration (API key, base URL)

### Phase 2: Firecrawl API Client
- [ ] Async HTTP client (httpx)
- [ ] Scrape/crawl/map/search/extract endpoints
[...]

PR #4682, opened 2026-08-21
feat(search): open grep + index to DT_STREAM content (nexus-vfs #229 P3)

elfenlieds7:
Opens the Rust search-plugin to **DT_STREAM** content — the last piece of nexus-vfs #229 P3 ("open search to streams"). A DT_STREAM (durable append-only log: audit, A2A mailbox, session transcript) now greps and indexes as one searchable document, exactly like a file.

## Changes
- **Plugin** (`search-plugin`): add `DT_STREAM=4` to `kernel_io` and a single SSOT predicate `searchable_content_type() = DT_REG || DT_STREAM`, routing all 5 VFS-walk gates (grep + keyword-index rec/non-rec + semantic-index rec/non-rec) through it. No per-site logic — DT_DIR/DT_MOUNT stay containers, DT_PIPE stays ephemeral.
- **Host enabler** (nexus-vfs, already merged in #235): the plugin C-ABI `sys_read` — a bulk, offset-less "content at path" read — now returns a DT_STREAM's *whole collected log* instead of just frame 0, so the plugin's existing `sys_read` call gets the full log with no plugin-abi change.
- **Pin bump**: nexus-vfs `c9ecc9da → b98c08e62` (to include the host shim). The 104-commit bump's only API drift was `sys_readdir` gaining a `ReaddirOpts` param; all 18 `services/` call sites now pass `ReaddirOpts::default()` (behavior-preserving). `cargo check --workspace` green.

## Live e2e (the honored validation)
`TestSearchStreams` in `tests/e2e/docker/test_search_plugin.py` (real cluster + signed plugin via the existing `docker-compose.search-plugin-e2e.yml`): seeds a wal stream whose token sits in a **LATE frame** (6th of 7), plus a sibling file, then —
- **grep** finds the late-frame token in the stream (proves the host read the *whole log*, not frame 0), the early-frame token, and the file (no regression); a bogus token matches nothing;
- **keyword index+query** surfaces the stream's content as a document;
- **semantic query** too (tolerant of a missing embedder).
Plus `vfs_create_stream` / `vfs_stream_write` runbook helpers.

Locally verified: plugin dev-signs + loads into a Phase-0 cluster, SearchService routes, wal-stream writes succeed, and the nexus workspace builds against the new pin. The stream grep/query assertions run in the Docker e2e (Linux, matched host+plugin, real signing key) — a Windows cross-DLL-allocator FFI issue (pre-existing, all plugin `sys_read`, not this change) blocks the same run locally on Windows.


elfenlieds7:
**⚠️ Blocked on the nexus-vfs → sudocode → nexus release chain — draft until resolved.**

The plugin change + `TestSearchStreams` are ready and the workspace builds against the new pin (only drift was `sys_readdir` gaining `ReaddirOpts`, adapted at all 18 `services/` sites). But the pin bump can't pass `test.yml`'s pin-consistency gate: it requires **every** nexus-vfs pin in Cargo.lock to agree, and the transitive `runtime` crate from `sudoprivacy/sudocode` (rev `d7f64b98…`) still pins nexus-vfs at the old `c9ecc9da`. So Cargo.lock has both `b98c08e62` (this repo) and `c9ecc9da` (sudocode) → 'pins diverge'.

**To unblock:** sudocode must bump its nexus-vfs pin to `b98c08e62` (which has #235's `sys_read`-collects-DT_STREAM host shim) and cut a release, then this PR re-pins sudocode to that rev. The nexus-vfs half is already merged (nexus-vfs#235, `b98c08e62`).

Issue #241, opened 2025-10-23
Ensure Identical Behavior Across Nexus Deployment Modes and Interfaces for Seamless Agent Integration


joezhoujinjing:
# GitHub Issue: Behavior Consistency Across Nexus Components

## Issue Title
**Ensure Identical Behavior Across Nexus Deployment Modes and Interfaces for Seamless Agent Integration**

## Issue Labels
- `enhancement`
- `architecture`
- `agent-integration`
- `consistency`

---

## Problem Statement

As an agent developer using the Nexus filesystem, I need **identical behavior** across all deployment modes and interfaces so that:

1. **Local backend and GCS backend** behave identically when using `nexus cli` commands
2. **Embedded Nexus and Remote Nexus** provide the same functionality and guarantees
3. **Nexus CLI commands** match the server RPC API methods exactly
4. **Agent tool calls** can directly map to Nexus CLI operations, enabling React agents to use Nexus out of the box

Currently, there are several behavioral inconsistencies and missing features that prevent seamless agent integration.

---

### 1.1 Local vs GCS Backend Behavioral Differences

**Current State:**

| Behavior | Local Backend | GCS Backend | Impact |
|----------|---------------|-------------|---------|
| **Concurrent Writes** | File locking with fcntl/msvcrt | No locking mechanism | **HIGH** - Race conditions possible |
| **Directory Implementation** | Native filesystem | Marker blobs with `/` suffix | Medium - Mostly transparent |
| **Empty Directory Cleanup** | Automatic cleanup | Manual via `rmdir` | Low - Different cleanup behavior |
| **Large File Timeout** | No timeout | 60-second timeout | Medium - Large files may fail on GCS |
| **Atomic Operations** | Temp + move pattern | Direct GCS upload | Low - Both work but different |

**Files Affected:**
- [src/nexus/backends/local.py](https://github.com/nexi-lab/nexus/blob/main/src/nexus/backends/local.py) (Lines 147-189 - write_content with locking)
- [src/nexus/backends/gcs.py](https://github.com/nexi-lab/nexus/blob/main/src/nexus/backends/gcs.py) (Lines 170-201 - write_content without locking)

**Specific Issue:**

```python
# Local backend (local.py:147-189)
def write_content(self, content: bytes) -> str:
    content_hash = compute_hash(content)
    content_path = self._hash_to_path(content_hash)

    with FileLock(content_path):  # ✓ Thread-safe
        if content_path.exists():
[...]

PR #4698, opened 2026-08-23
chore(deps): lockstep bump nexus-vfs d3b182951 + cohost sudocode f5b15e97

elfenlieds7:
## What

Lockstep bump of the two pins that must move together for the in-process
co-host seam:

- **workspace nexus-vfs pin** `6cad75f7b` → `d3b182951` (current nexus-vfs `main`)
- **nexusd co-host `sudocode` pin** `4bc7cba6` → `f5b15e97` (sudocode #482, which
  pins the SAME nexus-vfs rev `d3b182951`)

## Why

`d3b182951` restores two federation clean-slate features to develop that were
dropped when the pin last moved to a nexus-vfs `main` tip that lacked them
(they had only ever lived on a side branch):

- `reset` subcommand — one command for a truly-fresh node (clears data-dir AND
  identity.json), the fix for the phantom-voter / auto-rejoin-stale deadlock.
- raft fail-loud when an unreachable voter blocks quorum (diagnostic; names the
  dead voter + the fix).

Both are directly relevant to the cross-org / Win↔Mac co-host duet. It also
pulls #238 (`sys_stat` DT_STREAM last-append time).

The co-host `SpawnTask<Kernel>` seam links nexusd's own `kernel` (workspace pin)
and sudocode's `kernel` (git dep) — they MUST resolve to one rev or the type
won't unify. sudocode #482 bumped its co-host `kernel`/`a2a` pin to the same
`d3b182951`, so this bump keeps the seam single-typed.

## Test

- Lockfile regenerated; no residual old revs (`6cad75f7b` / `4bc7cba6` = 0).
- Default-feature build validated by CI here.
- The `--features cohost-sudocode` seam is validated by a Linux release build of
  the deploy binary (the on-VM cohost broker for the cross-org duet).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ version = 4
 [[package]]
 name = "a2a"
 version = "0.1.0"
-source = "git+https://github.com/nexi-lab/nexus-vfs?rev=6cad75f7b43e49b60c9a345521695e915a6d7b47#6cad75f7b43e49b60c9a345521695e915a6d7b47"
+source = "git+https://github.com/nexi-lab/nexus-vfs?rev=d3b182951569d22604e03b0a2b47e950fa96c77e#d3b182951569d22604e03b0a2b47e950fa96c77e"
 dependencies = [
  "contracts",
  "kernel",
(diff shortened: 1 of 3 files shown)


elfenlieds7:
Superseded: develop has already advanced (via #241/#4695 in the nexus-duet line) to nexus-vfs `dcee3e5d` + sudocode `155dd3a1`, both newer than this PR's `d3b182951`/`f5b15e97`. `dcee3e5d` already includes the reset+fail-loud restore (nexus-vfs #239), so this lockstep bump would only regress the pins. Closing.

Issue #4740, opened 2026-09-03
security: zone defaults fall open on list/search: missing zone resolves to the ROOT view, root-tagged entries visible from every zone, admin skips the zone post-filter


windoliver:
Observed at 9e7d6af (2026-09-02 code read).

## What happens

- `src/nexus/core/nexus_fs_metadata.py` `sys_readdir`: a **missing zone claim resolves to the ROOT global view "by design"** (1899-1904); **root-tagged entries pass `_zone_allowed` from every zone** (1957-1963); the recursive / detailed / paginated list path is a Python post-filter over the global metastore (1931-1963), and the comment at 1836-1840 says that path misses federation-zone entries the kernel readdir would return.
- `src/nexus/bricks/search/search_service.py:630-638`: the search `list` zone filter is a Python post-filter and is **skipped for admins**.
- `src/nexus/server/zone_execution.py:22-40` `context_for_target_zone` mutates the caller's context in place, granting admins `zone_perms=((target, "rw"),)`.
- The Rust `http-api` search/documents handlers carry `zone_id` but have no file-level ReBAC post-filter yet (`rust/services/http-api/src/handlers/search.rs`; `rust/services/rebac/src/lib.rs:23-26`, PR-4 pending).
- #3983's self-assessment already lists "SDK uses base ReBACManager without tenant isolation" as P0.

## Consequence for a tenant app

Anything written without `/zone/<id>/` scoping (embedded mode, an admin key, root-zone traffic) is listable by every tenant, and a caller that omits the zone header sees the global namespace instead of being refused. Lists and searches are otherwise post-filtered, so the usual failure mode is *missing* results, but the two root-zone paths above are *leaks*. Koodle currently sends no `X-Nexus-Zone-ID` and seeds `path_contexts` under zone `root`, so it is relying entirely on path prefixes plus its own Postgres RLS for isolation.

## Proposal

1. **Deny zone-less non-admin contexts** on list and search instead of resolving ROOT.
2. **Require a zone tag on every write**; root-tagged entries are visible only to root-zone callers.
3. Move the zone predicate into the metastore prefix scan for list (pre-filter) and into the plugin query for search (the P12 plugin already indexes per zone directory, so the search half is a routing check).
4. Admin listing across zones becomes an explicit, audited `all_zones=true`.
5. A **matrix test**: tenant A key lists and searches with and without the zone header and never sees tenant B or root-tagged entries; an anti-vacuity control as root sees both.

## Acceptance

[...]

Issue #341, opened 2025-10-31
ReBAC permissions not updated during rename/move operations


joezhoujinjing:
## Problem

When renaming or moving a file using the `rename` RPC method, the ReBAC permissions are not updated to reflect the new path. This causes files to become inaccessible after moving them.

## Steps to Reproduce

1. Create a file at `/workspace/user_a/test.txt`
2. Grant permissions: `rebac_create user user_a direct_editor file /workspace/user_a/test.txt`
3. Rename/move the file: `nx.rename('/workspace/user_a/test.txt', '/workspace/shared/test.txt')`
4. Try to access the file at the new location
5. **Expected**: File is accessible at `/workspace/shared/test.txt`
6. **Actual**: File is inaccessible because ReBAC still references old path

## Root Cause

The `rename_path()` method in `nexus/src/nexus/storage/metadata_store.py:589-628` only updates the `FilePathModel.virtual_path` field. It does not update any ReBAC tuples that reference the old path.

## Proposed Solution

When renaming a file, the backend should:
1. Find all ReBAC tuples where `object_type='file'` and `object_id=old_path`
2. Update those tuples to use `object_id=new_path`
3. For directories, also update all child paths recursively

## Workaround

Currently, there's no workaround. Users must manually:
1. Delete old ReBAC permissions
2. Re-create permissions for the new path

This is not feasible for nested directories with many permission rules.

## Impact

- **High**: Makes drag-and-drop file management unusable
- Breaks file organization workflows
- Causes data to become "lost" (still exists but inaccessible)

## Related Code

- `nexus/src/nexus/core/nexus_fs_core.py:847` - `rename()` method
- `nexus/src/nexus/storage/metadata_store.py:589` - `rename_path()` method

PR #4746, opened 2026-09-03
feat(rebac): sync owner grant, consistency=strong, targeted Tiger revocation, ReBAC-filtered readdir (#4739)

### LINKED ISSUE #4739: rebac: deferred owner grant by default + 60 s denial cache + 1 h Tiger cache: a non-admin writer can be denied its own file; add sync owner grant and consistency=strong
Observed at 9e7d6af (2026-09-02 code read).

## What happens

- `src/nexus/core/config.py:61` `enable_deferred=True` selects `DeferredPermissionHook` (`bricks/rebac/deferred_permission_hook.py:57-70`), which **queues** the hierarchy tuples and the creator's owner grant after a write, flushed every 100 ms (`deferred_permission_buffer.py:47`). `SyncPermissionWriteHook` exists (`sync_permission_hook.py:1-12`) but is not the default.
- `src/nexus/cache/settings.py:27-36, 74-106`: grants cached 300 s, **denials 60 s**, Tiger bitmaps 3600 s, owner 1 h / editor+viewer 10 min / inherited 5 min. `rebac_service.py:440` and `manager.py:502,875`: "Always uses cached (eventual) consistency". `permission_lease.py` adds a TTL lease ("check once, write many").
- #1071 introduced the async buffer for write throughput; the consistency window it created is not documented.

## Consequence for a tenant app

With a non-admin key: write returns 200; the writer's immediate `read` / `list` / `search` can be denied or filtered until flush + cache invalidation; a denial cached inside that window persists 60 s; a revoked viewer can keep seeing a file for up to 3600 s via the Tiger bitmap. Today Koodle runs hot paths on an admin key (bypass), so this is latent; it becomes live the moment per-team non-admin keys carry hot-path traffic (koodle#143 residual), which is the direction both projects want.

## Proposal

1. The **owner grant for the creating subject is written synchronously** (or a permission lease is stamped at write time for the creator); hierarchy tuples may remain deferred.
2. `consistency=strong` on `check`, `list`, and `search` bypasses L1 and Tiger for that call (SpiceDB `fully_consistent` analog).
3. **Tuple delete triggers targeted Tiger invalidation** (`coordinator.py:538`) so revocation is bounded by seconds, not the 3600 s TTL.
4. Document the cache matrix (what is cached, TTL, invalidation trigger) in the ReBAC docs.

## Acceptance

- Test: non-admin key writes, then reads, lists and searches within 10 ms and sees the file.
- Test: revoke a viewer, `check` within 1 s denies; `list` and `search` exclude the file within 1 s.

## Filed alongside (2026-09-02 read)

nexi-lab/nexus: #4736 write-to-searchable contract · #4737 read-your-writes revision token · #4738 fenced projections · #4739 sync owner grant + consistency=strong · #4740 zone fail-open on list/search · #4741 read-after-write conformance suite.
Tenant-side tracking: DeepBuildAI/koodle#3079 (Nexus compensation ledger; these six are its retire conditions).

windoliver:
## Summary

Closes #4739.

A non-admin writer could be denied its own file and a revoked viewer could keep seeing files for up to the Tiger TTL. This PR implements the four proposals in the issue and fixes two prerequisites found on the way.

- **Sync owner grant.** `DeferredPermissionHook` writes the creator's `direct_owner` tuple synchronously (`rebac_write`, `rebac_write_batch` for batch writes) and defers only the hierarchy tuples. Failure falls back to the queue with a `degraded` warning. New knob `PermissionConfig.sync_owner_grant` (default on) restores the old behaviour when off.
- **Kernel client `is_new`.** The typed `WriteResponse` carries only `content_id/size/gen`, and the Python client hard-coded `is_new=False`, so no post-write hook ever saw a new file over the kernel RPC and the owner grant (deferred or sync) never fired. `is_new` is now derived as `gen == 1`, which is the kernel's own `old_entry.is_none()` condition.
- **`consistency=strong`.** New `OperationContext.consistency`, set from `X-Nexus-Consistency` (400 on invalid values). `rebac_check`, `rebac_check_bulk`, `rebac_check_batch(_fast)`, the enforcer check and `filter_list`, search's Tiger predicate pushdown, the lease fast path and directory visibility all bypass cached decisions for that call and write the fresh result back. `rebac_check` / `rebac_check_batch` RPCs take an explicit `consistency` parameter. Also fixes `AsyncReBACManager.rebac_check` passing a sixth positional argument the sync method did not accept.
- **Targeted Tiger invalidation on delete.** `rebac_delete` routes through `CacheCoordinator.invalidate_for_write` (leases, Tiger L1/L2 eviction, cross-process hints) and, for directory grants, drops the grant record and the subject's bitmap so Leopard-expanded descendants do not stay visible for 3600 s. Grants keep the cheap write-through path.
- **L1 polarity.** A fresh denial now evicts a stale cached grant for the same key (the grant cache was consulted first, so a stale allow outlived a fresh deny).
- **`sys_readdir` ReBAC filter.** The raw syscall (behind `/v2/files/list`) applied zone filtering only; non-admin, non-system callers now get the same file filter as search list/glob/grep plus descendant-based directory visibility. Admin/system/identity-less callers are unchanged (#4740 covers admin zone scoping). Fails closed if enforcement is on but the enforcer is missing.
- **Distributed lease invalidator.** The lifespan callback took one argument but the coordinator passes four, so it raised on every notification and never revoked distributed leases; fixed with a contract test.
- **RPC params regenerated** (`--check` is clean again; the drift was stale snapshot params).
- **Docs.** `docs/architecture/rebac.md`: cache matrix (layer, TTL, writer, invalidation trigger, strong bypass), sync owner grant, revocation bound, listing surfaces.

## Test plan

- [x] Unit: `tests/unit/bricks/rebac/{test_deferred_permission_hook,test_enforcer_strong_consistency,test_consistency_strong,test_permission_filter_chain}.py`, `tests/unit/core/test_readdir_rebac_filter.py`, `tests/unit/remote/test_kernel_client_write_result.py`, `tests/unit/server/{test_consistency_header,test_distributed_lease_invalidator}.py`; `tests/unit/{core,bricks/rebac,services/permissions,remote,server,lib,bricks/search}` green. ruff / mypy clean on changed files.
- [x] Acceptance (kernel-backed, SQLite): `tests/integration/bricks/rebac/test_read_after_write_permissions.py` — non-admin write then immediate check/read/list/glob/grep with the deferred buffer never flushing; owner tuple stands without ancestor grants; revoked viewer denied and excluded from list/glob/grep and raw `sys_readdir` immediately; strong context honoured end to end; legacy window reproduced with `sync_owner_grant=False`. Run with `NEXUS_BOOTSTRAP_MODE=static` on the local kernel binary.
- [x] Live on `nexus up --build` (Postgres + Dragonfly + Rust search plugin host, Tiger active): 37/37 HTTP checks (write → immediate `rebac_check`, read, list, glob, grep, search as the writer; owner tuple after revoking the directory grant; viewer revoke denied on the first `rebac_check` and excluded from list/glob/grep/search; `X-Nexus-Consistency: strong` on list/glob/search, `bogus` → 400; `rebac_check` RPC with `consistency`); 24/24 for a revoked *directory* viewer grant with Leopard expansion (nested file denied on the first check, list/glob/grep excluded).
- [x] `permissions_demo_enhanced.sh` inside the container: all tests passed.
- [x] `scripts/test_build_perf_e2e.py` inside the container: 30 passed, 0 failed, HERB 8/8, `backend=rust-plugin`.
- [ ] Both scripts re-running on the image rebuilt with the lease-invalidator fix (server log shows zero `Lease invalidator … failed` after revokes); result will be posted as a comment.

## Follow-ups (not in this PR)

- nexus-vfs: `WriteResponse` should carry `is_new` and the `old_*` fields; the kernel does not stamp `FileMetadata.owner_id` on write, so the owner fast-path is inert over the RPC.
- Zone-isolated single-check denials are not served from L1 (`rebac_check_detailed` never consults it) and each fetches tuples per check: 300–900 ms per denial live, one `GraphLimitExceeded` timeout (>1 s) observed on a `/files/read` denial. Pre-existing; not changed here.
- `rebac_check_batch_fast` still ignores `zone_id` (pre-existing).

### DIFF
--- a/docs/architecture/rebac.md
+++ b/docs/architecture/rebac.md
@@ -246,3 +246,112 @@ self.rebac = EnhancedReBACManager(engine)
 | **Our changes apply to** | ✅ | ✅ (inherited) | ✅ (inherited) |
 
 **Bottom line:** Our changes to `ReBACManager` automatically improve all three classes! 🎉
+
+---
+
+## Permission caches, TTLs and consistency (Issue #4739)
+
+A permission decision (`check`, and the `filter_list` pass behind `list`,
+`glob`, `grep` and search) can be answered by several caches before the tuple
+store is consulted. Each cache has its own TTL and its own invalidation
+trigger, so the window between a tuple change and every surface reflecting it
+is the *longest* of the rows below unless the caller opts into
+`consistency=strong`.
+
+### Cache matrix
+
+| Layer | What it caches | Default TTL (env) | Written by | Invalidated by | Bypassed by `consistency=strong` |
+|---|---|---|---|---|---|
+| **Owner fast-path** (`PermissionEnforcer.check_owner`) | Nothing cached — compares `metadata.owner_id` to the subject on single-path `read`/`write`/`delete` | n/a | Kernel stamps `owner_id` on create | n/a | No (authoritative) |
+| **Permission lease** (`PermissionLeaseTable`) | A successful single-path check for `(path, agent_id)`; only agents (`context.agent_id`) get leases | 30 s | `RebacPermissionCheckHook` after a full check | `rebac_delete` → `CacheCoordinator.invalidate_for_write` (path-targeted for direct grants, zone-wide for group/inherited); agent termination | Yes |
+| **L1 result cache** (`ReBACPermissionCache`) | `(subject, permission, object, zone) → bool` | grants 300 s `NEXUS_CACHE_PERMISSION_TTL`; denials 60 s `NEXUS_CACHE_DENIAL_TTL`; tiered: owner 3600 s, editor/viewer 600 s, inherited 300 s | `rebac_check` (base path), `rebac_check_bulk`, eager recompute | Every `rebac_write` / `rebac_delete` for the touched subject and object (plus object prefix for file relations) | Yes |
+| **Boundary cache** (`PermissionBoundaryCache`) | Nearest ancestor that granted `permission` on a file path | process lifetime, revalidated against a direct-grant lookup on hit | `_check_rebac_*` when a parent granted | `rebac_write` / `rebac_delete` via boundary invalidators | Yes |
+| **Tiger bitmap** (`TigerCache`, PostgreSQL only) | Per `(subject, permission, resource_type)` roaring bitmap of accessible resources; L1 in-process, L2 Dragonfly, L3 table | 3600 s `NEXUS_CACHE_TIGER_TTL` (Dragonfly key TTL; L3 rows have none) | Write-through on `rebac_write` (`persist_single_grant`, Leopard-style directory expansion to descendants), background updater | `rebac_delete`: `persist_single_revoke` for the exact object; for a **directory** grant also `remove_directory_grant` + `tiger_invalidate_cache(subject, permission)` (drops L1+L2+L3); `invalidate_for_write` evicts L1/L2 for the subject in every process | Yes (`check`, `filter_list`, search predicate pushdown, directory visibility) |
+| **Bitmap completeness / Leopard dir index** (`PermissionCacheCoordinator`) | "This subject's bitmap is complete" and accessible-directory sets used by `filter_list` | 3600 s | `BulkReBACStrategy`, `HierarchyPreFilterStrategy` | `PermissionCacheCoordinator.invalidate` | Yes (strong chain skips both strategies) |
+| **Zone graph cache** | Tuple graph snapshot per zone for the Rust evaluator | until next write | graph loader | Every `rebac_write` / `rebac_delete` in the zone | n/a (always refreshed on write) |
+| **Deferred permission buffer** (`DeferredPermissionBuffer`) | Not a cache: hierarchy (`parent`) tuples waiting to be written | flushed every `PermissionConfig.deferred_flush_interval` (50 ms) or at 1000 items | `DeferredPermissionHook` on write/mkdir/write_batch | n/a | n/a — see "sync owner grant" |
+
+### Sync owner grant (read-your-writes for the creator)
+
+With `PermissionConfig.enable_deferred=True` (default) the hierarchy tuples for
+a new file are still batched, but since #4739 the creating subject's
+`direct_owner` tuple is written **synchronously** by `DeferredPermissionHook`
+(`rebac_write`, or `rebac_write_batch` for `write_batch`). The writer's own
+`check`, `list`, `glob`, `grep` and search therefore see the file as soon as
+the write returns; no flush or cache expiry is involved. If the synchronous
+write fails, the grant falls back to the deferred queue and the write result
+carries a `degraded` warning (`component=deferred_permission`).
+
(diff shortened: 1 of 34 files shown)


windoliver:
## Live verification on `nexus up --build` (Postgres + Dragonfly + Rust search plugin host, Tiger active)

Stack from this worktree (ports 12660-12664), plugin host on 2127, scripts run inside the container like CI (`NEXUS_URL=http://127.0.0.1:2026`, admin + demo-user keys, `nexus reindex --target search` first).

| Pass | Image | `permissions_demo_enhanced.sh` | `scripts/test_build_perf_e2e.py` |
|---|---|---|---|
| 1 | branch before rebase | all tests passed | 30 passed, 0 failed, HERB 8/8, `backend=rust-plugin` |
| 2 | + distributed-lease invalidator fix | all tests passed | 21/30 — every failure was a search endpoint 500 because the plugin host process on 2127 had been killed externally mid-run (server health: `connection refused` to 2127); not a code failure |
| 3 | rebased on `develop` (+ #4744, #4745) + readdir-filter module | exit 0, 2 non-blocking warnings (the known #341 zone-scoped rename case the script tolerates) | 30 passed, 0 failed, HERB 8/8 |

Feature checks over HTTP with freshly minted non-admin keys (`live_4739_check.py`, 36 checks; `live_4739_dirgrant.py`, 25 checks) — all pass on the pass-3 image:

- write as a non-admin, then immediately: `rebac_check` True, read 200, `/files/list`, glob, grep include the file; the whole sequence in ~450 ms with the deferred buffer untouched.
- owner tuple stands alone: after revoking the writer's directory grant, check/read/list still succeed.
- viewer revoke: first `rebac_check` after `DELETE /rebac/tuples` is denied (0.15–0.3 s incl. HTTP), read 403, list/glob/grep/search exclude the file.
- **directory** viewer revoke with Leopard expansion: nested file denied 0.15 s after the revoke; list/glob/grep exclude all three files (this is the 3600 s Tiger case from the issue).
- `X-Nexus-Consistency: strong` on list/glob/search works; `bogus` → 400; `rebac_check` RPC accepts `consistency=strong|eventual`.
- rename probe (root zone): both tuples follow the rename, old path denied, new path allowed.
- server log after the fix: 0 `Lease invalidator … failed`, 0 `TypeError`, 0 `GraphLimitExceeded` (one `GraphLimitExceeded` timeout was seen on a denial path on the pass-1 image; denial checks cost 300–900 ms on this stack — pre-existing, see follow-ups).

Local unit run after the rebase: `tests/unit/{core,bricks/rebac,remote,services/permissions,lib,server,bricks/search}` green except `tests/unit/remote/test_grpc_target.py::test_default_port_no_tls`, which reads the worktree's gitignored `nexus.yaml` created by `nexus init` (environmental; CI has no such file). Kernel-backed acceptance suite `tests/integration/bricks/rebac/test_read_after_write_permissions.py`: 6 passed.

PR #4748, opened 2026-09-04
perf(write path): admin-bypass owner-grant skip, NEXUS_PROJECTION_MODE=async, NEXUS_SYNC_OWNER_GRANT; fix(search-plugin): lenient query parse

windoliver:
## Problem

Validating develop `327dfd1` (PRs #4742–#4746) against a production-shaped deployment (Railway us-west2, paradedb pg18 in-region, Dragonfly, Rust search plugin; every write made with an admin key under `NEXUS_ALLOW_ADMIN_BYPASS=true`) showed a write-path regression with no way to opt out. Reads, list and search were unchanged; retrieval quality, the SDK surface (23/23 parity), and in-place upgrade/rollback were all clean.

| interleaved A/B, client p50 | old `f229e4b` | new `327dfd1` |
|---|---|---|
| `POST /files/write`, Railway in-region | 105–133 ms | 320 ms |
| `POST /files/batch/write` (3 files), Railway | 101–116 ms | 506 ms |
| server-side `files/write` `duration_ms`, Railway | 24–30 ms | 200–260 ms |
| `files/write`, local docker | 29–38 ms | 76–98 ms |

A Postgres statement log (`log_min_duration_statement=0`) of one new-file write on each build shows the two causes:

1. **#4739 sync owner grant for admin-bypass writers.** `KernelClient` now derives `is_new = gen == 1`, so the post-write hook fires and `DeferredPermissionHook._owner_grant` does a synchronous `rebac_write(direct_owner)`: one `tiger_resource_map` insert, three `tiger_cache` inserts (read/write/execute), the tuple row, a zone-graph invalidation — ~8 extra commits per new file. Before #4739 the subprocess-kernel deployment never wrote these (`is_new` was hard-coded False), and under admin bypass the tuple is never consulted. A 150-write probe left 453 `direct_owner` rows vs 3. `PermissionConfig.sync_owner_grant` existed but was code-only.
2. **#4738 write-through projections.** The writer now blocks on the `operation_log` / `file_paths` / `version_history` transaction (~10 round-trips + fsync). Skipping the owner grant alone left Railway writes at 191 ms vs 105 ms (server-side 100–130 ms vs 24–30 ms).

## Change

[...]

PR #4749, opened 2026-09-04
fix(search-plugin): natural-language queries fall back to tantivy's lenient parser

windoliver:
## Problem

Running the LongMemEval retrieval benchmark (`longmemeval_s`, cleaned Sept-2025, the protocol from garrytan/gbrain-evals) against a Nexus server built from develop `327dfd1`, question `58470ed2` ("… and I wanted to confirm - what did Borges say about the center and circumference of the Library?") failed on **every** keyword and hybrid fusion mode with

```
search: parse query "…": Syntax Error
```

while the semantic lane had the gold session at rank 1. `FtsIndex::search` parses the query strictly with tantivy's `QueryParser::parse_query`; a dash surrounded by spaces, an unbalanced quote, a stray `:` or `[` — all things a natural-language question carries by accident — reject the whole query, and because the hybrid path joins the keyword leg with the semantic leg, the whole hybrid RPC fails. Koodle works around this today with a client-side sanitizer in its SDK; any other client (CLI, MCP, the Python SDK) gets the 500.

## Change

Try the strict parser first so deliberate operator syntax keeps its exact semantics; on a syntax error fall back to `QueryParser::parse_query_lenient`, which drops the offending tokens and searches the remaining terms. Strict error + lenient error count are logged at debug.

## Verification

| Check | Result |
|---|---|
| `natural_language_syntax_falls_back_to_lenient_parse` (new): three accidental-syntax questions hit the right doc; `Borges -gardening` still excludes the negated term | pass |
| `cargo test -p nexus-search-plugin` (267 unit + integration binaries), `cargo clippy --all-targets -D warnings`, `cargo fmt --check` | pass |
| Live: plugin rebuilt, dev-signed and hot-swapped into a `nexusd-cluster --plugin-dir` host behind a `nexus up --build` stack; the 50-session haystack of `58470ed2` indexed, raw question queried | keyword hits=15 gold rank 1, hybrid hits=15 gold rank 1, semantic gold rank 1 (before: keyword + every hybrid mode → Syntax Error) |

Independent of #4748 (write path); both apply cleanly on develop.

### DIFF
--- a/rust/services/search-plugin/src/fts_index.rs
+++ b/rust/services/search-plugin/src/fts_index.rs
@@ -576,9 +576,29 @@ impl FtsIndex {
         let searcher = self.reader.searcher();
 
         let parser = QueryParser::for_index(&self.index, vec![self.fields.chunk_text]);
-        let parsed = parser
-            .parse_query(query)
-            .map_err(|e| IndexError::ParseQuery(query.to_string(), e.to_string()))?;
+        // Natural-language queries routinely contain tantivy syntax by
+        // accident ("confirm - what did ...", an unbalanced quote, a
+        // stray ':' or '[').  The strict parser rejects the whole query,
+        // and because the hybrid path joins this leg with the semantic
+        // leg, one dash used to fail the entire hybrid RPC with
+        // "parse query ...: Syntax Error" even though the ANN leg had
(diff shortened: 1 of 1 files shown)


windoliver:
Folded into #4748 (commit 06dab20 on that branch) so the whole thread ships as one PR.

PR #4786, opened 2026-09-13
chore(deps): bump nexus-vfs pin to 62bc2f5b — mTLS arms identity

elfenlieds7:
Pin moves `35df7535` (nexus-vfs#273) → `62bc2f5b` (main tip). Two-PR delta, nothing incidental.

## What it brings

**nexus-vfs#275 — mTLS arms identity, not just admission.** `auth_posture::decide()` returned `Open` for a reachable bind with TLS and no `NEXUS_API_KEY_SECRET`, on the reasoning that a verified client certificate is itself an authentication. It is — for admission. But `Open` installs no provider that reads the certificate, so `auth_armed` was false, the A2A stamp hook ran fail-open, and an envelope's authored `from` passed through unstamped. There is now a certificate-identity posture whose provider resolves the cert (SAN → agent id, foreign orgs org-qualified, CRL honoured) and rejects a caller presenting none.

Scope worth stating plainly: a founder self-provisions the cluster secret and enrollment writes it to a joiner, so both of those were always on the `sk-` posture and were never affected. What this fixes is the node whose certificates arrive **out of band** — a customer PKI, a hand-built fixture.

**nexus-vfs#274** — the version-stamp contract doc (`NEXUSD_BUILD_VERSION` carries the version, not the ref name).

## Why the Dockerfiles move in the same commit

`Cargo.toml`, `Cargo.lock` and the three `NEXUS_VFS_REV` ARGs (`Dockerfile.minimal`, `.raft-server`, `.raft-witness`) all name the same rev. A half-bump builds an image against a different kernel than the workspace resolves, which is the kind of skew that shows up later as an unexplained type or ABI mismatch rather than as a build failure here.

## Verification

`cargo check --workspace` clean against the new rev (Windows, full MSVC env). CI carries the rest.


elfenlieds7:
**Blocked — this bump is a two-repo chain, and the pin-consistency gate is right to refuse it.**

`nexusd-cluster Smoke Test` fails with:

```
Cargo.lock pins:  35df753583d15aa26c14b5ac4885ec1a210eb2d4
                  62bc2f5b45a31a36a8a17e41ab605e32d4e20218
Dockerfile pins:  62bc2f5b45a31a36a8a17e41ab605e32d4e20218
error: nexus-vfs pins diverge across Cargo.toml / Cargo.lock / Dockerfiles — partial bump ships an untested kernel
```

The second rev is not a missed file. `sudocode` (`runtime`, `tools` at `66abb4af`) pins nexus-vfs at the **old** `35df7535`, so moving only the workspace pin resolves two kernels in one lock — exactly what the gate exists to stop, and what would make `SpawnTask<Kernel>` two distinct types in a `cohost-sudocode` build.

Order that works:
1. **sudocode** bumps its nexus-vfs pin `35df7535` → `62bc2f5b`;
2. **here**, bump the nexus-vfs pin *and* the sudocode pin in the same commit (plus the three Dockerfile ARGs, as this branch already does).

Leaving this open as the second half. What it is waiting to deliver: nexus-vfs#275, where a daemon whose only credential plane is mTLS installs a provider that actually reads the certificate — before it, such a node admitted only CA-verified callers and then resolved every one as nobody-in-particular, leaving A2A `from` stamping fail-open. Out-of-band certificates (a customer PKI) is precisely that shape.

PR #4799, opened 2026-09-17
Add Linux aarch64 vault release artifacts

yinbinchen:
## Summary
- add linux-aarch64 to the reusable plugin release matrix
- include linux-aarch64 in the vault release workflow so both nexus-vault and nexusd-vault tarballs are produced
- add linux-aarch64 to the vault plugin build validation matrix

## Validation
- ruby YAML parse for the touched workflows
- git diff --check
- actionlint on the touched workflows


elfenlieds7 (approved):
LGTM — the matrix addition is correctly gated and backward compatible.

Checked the part a diff does not show, i.e. whether anything still assumes four platforms: in `release-plugin-reusable.yml` every step carries `if: contains(fromJSON(inputs.platforms), matrix.artifact_name_suffix)` (lines 165/169/186/218/230/237/243/266/281/300/315), so the new matrix entry is gated like the rest and no step silently runs or silently skips. The COS path is parameterised by `artifact_name_suffix`, and the size budgets live per matrix entry, with aarch64 given the same 6 MiB as linux-x86_64. `default` still lists the original four, so plugins that do not opt in are unaffected.

One thing worth knowing rather than fixing here: this workflow cannot run on a PR, so YAML parse + actionlint is as far as the gate reaches — the aarch64 leg has not actually built. I would not dispatch it just to find out, since a real run publishes to COS and cuts a Release. Suggest watching that leg on the first vault release after this lands; if `ubuntu-22.04-arm` needs anything extra (cross deps, protoc), it will surface there and nowhere earlier.

